PR #4, opened 2013-06-12
retry loading cross context component metadata from viewer-admin if viewer not deployed

matthijsln:


matthijsln:
To test: add this to log4j.properties in viewer-admin:

log4j.logger.nl.b3p.viewer.components=DEBUG

This change should enable the following to work:
- deploy viewer-admin first (no components can be loaded)
- when you view the layout of an application, no components are loaded
- deploy viewer
- view the layout of the application again, components should be loaded from viewer

mtoonen:
Nice addition!

PR #72, opened 2013-10-16
Feat/application layer context

geertplaisier:
Added LayerContext component and added 'Context' config tab for application layers to solve issue #57


mtoonen:
Seems ok. Tested and works. Merging. Thanks @geertplaisier

PR #83, opened 2013-10-31
Set Waiting cursor when loading Feature info. Implemented for OpenLayers...

rbraam:
.... Need to be implemented for Flamingo


rbraam:
Is not working for IE... see: #88

PR #201, opened 2014-05-20
First load data, after that, load the count. This to enhance user experi...

mtoonen:
...ence: loading data is relatively quick with databases and counting relatively slow.


mtoonen:
@matthijsln Is this smart? Do you know if this would have a negative impact on WFS and arc\* servers? I imagine for WFS services that the count is implied from the number of results (in WFS 1.1 there is no paging, so numFeatures== totalCount), so splitting these operations might have a negative impact on user experience (it takes twice as long). Ideas?

PR #211, opened 2014-05-27
Dbk print xsl

escheper:
The default print xsl templates are replaced and new templates are added to support maps with DBK info.


mtoonen:
Specific components stuff not in general repo.

PR #213, opened 2014-06-03
Feat/grouped components

mtoonen:
What this change does, is in the viewer-admin->layoutmanager it groups the components, instead of showing one terribly long list.
A group of components is shown in a collapsible panel. Groups can be defined in the metadata in components.json using the property "group" with the value.


mtoonen:
#217 must be fixed before merging

PR #226, opened 2014-06-19
Feat/attributelist download dataset

mtoonen:
This PR introduces an extension for the attributelist component. It adds a dropdown and a button, so the user can specify which outputformat is wanted for downloading the current selection of features. 
It takes into account the current filter.


matthijsln:
Is the temporary file deleted?

PR #228, opened 2014-06-20
Feat/print extra information

mtoonen:
This change adds the functionality to add extra information to a print.
For information how to use:
https://gist.github.com/mtoonen/5b97b82255b8e6ae39ac

### DIFF
--- a/viewer/src/main/java/nl/b3p/viewer/stripes/PrintActionBean.java
+++ b/viewer/src/main/java/nl/b3p/viewer/stripes/PrintActionBean.java
@@ -14,6 +14,9 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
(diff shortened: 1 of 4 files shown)


viewer/src/main/java/nl/b3p/viewer/stripes/PrintActionBean.java
```diff
@@ -152,6 +167,29 @@ public Resolution print() throws JSONException, Exception {
...
+            JSONArray jarray=null;
+            Object o = jRequest.get("extra");
+            if (o instanceof JSONArray){
+                jarray= (JSONArray)o;
+            }else if (o instanceof String){
+                jarray = new JSONArray();
+                jarray.put(o);
+            }
+            for (int i=0; i < jarray.length();i++){
+                String extraStr = jarray.getString(i);
```
matthijsln: Seems to me this wouldn't work as the JSONArray does not contain strings but objects (see line 696)

PR #231, opened 2014-06-24
Two bugfixes:

mtoonen:
1. Authorization modified date was not updated correctly
2. Check if cache was valid, was incorrect: didn't include the equals case
   Fixes #230


matthijsln:
Thanks for fixing!

PR #384, opened 2015-07-02
Prevent geometry editing by specified user groups on layers with editable geometry

mprins:
This makes it possible to have an editable layer that can be fully editable for one or more groups and have one or more groups that cannot edit the geometry attribute (but can edit the other editable attributes) of the layer.

Not being able to edit the geometry also prevents deleting, splitting or merging the feature.


mprins:
@mtoonen if you have some time...

image

PR #389, opened 2015-07-07
Add a handle to add extra links to the popup

mprins:
This will make it possible to add extra action links as a callback to the featureinfo (or maptip) summary by calling `registerExtraLink: function (component, callback, label)`


mprins:
@mtoonen Maybe I need to add a filter for appLayer to this as well? so it doesn't render the link for all layers that the Maptip / FeatureInfo applies to

PR #400, opened 2015-08-11
use a system chosen, ephemeral port to bind to for SearchClientTest

mprins:
nl.b3p.viewer.search.SearchClientTest fails on AppVeyor with the following stack trace:

```
nl.b3p.viewer.search.SearchClientTest
java.net.BindException: Address already in use: bind 
    at sun.nio.ch.Net.bind0(Native Method)
    at sun.nio.ch.Net.bind(Net.java:444)
    at sun.nio.ch.Net.bind(Net.java:436)
...
```

it tries to start a server on port 8888 which seems to be unavailable at some times.


mprins:
:hand: specifically the SearchClientTest should pass on AppVeyor before merging this

PR #405, opened 2015-08-31
fix the overview map to always have the same dimensions

mprins:
and not depend on the size of the viewport.

_NOTE_: this change may have effect on existing templates, especially if you used cm. to set the overview size.


mtoonen:
@escheper This also affects the dbk-stylesheets

PR #421, opened 2015-09-15
Add an 'Edit copy of existing feature' button

mprins:


mprins:
@mtoonen Any comments on this? 
Adds a 'Kopie' button to edit a copy of a feature. The button can be configured (on/off) from the viewer-admin similar to 'Verwijderen'

PR #431, opened 2015-09-23
Keep the min/max/close buttons on top of the popup

mprins:
by adding them to the balloon instead of to the balloon content.

Now the buttons will remain visible when the content is larger than the popup and it is scrolled where before the buttons would disappear.


mprins:
@geertplaisier Please review and merge when OK

geertplaisier:
Fixed this in a different way. Pushed fix to master.

334214c3f09cf31441e8b912f2667f46d59aac8a

PR #444, opened 2015-10-13
Change getter methods for accessing properties directly

### LINKED ISSUE #430: Overviewmap is squashed when using static image
Using an overview map based on a static image, the overview is squashed, it seems the image has width/heigth css attributes that are wrong (at least width is)

image

see: http://ibis.b3p.nl/app/IBIS2/vvjaap

I'm curious if someone touched css class `img.olTileImg`?

@mtoonen can you provide that link to the working version (luchtkwaliteit?)

@geertplaisier any ideas?

## for reproduction:

config param | value
------------ | -------------
Breedte | 150
Hoogte | 110
Positie | Rechtsonder
URL | http://flamingo.prvgld.nl/print/Overzichtskaart.png
lo-x | 123776
lo-y | 408506
rb-x | 258718
rb-y | 506174

mprins:
This seems to fix the aspect ration problem, but the overview is a bit off in the panel.

(sorry about the formatting changes)

close #430

### DIFF
--- a/viewer/src/main/webapp/viewer-html/common/viewercontroller/openlayers/components/OpenLayersOverview.js
+++ b/viewer/src/main/webapp/viewer-html/common/viewercontroller/openlayers/components/OpenLayersOverview.js
@@ -1,5 +1,5 @@
 /* 
- * Copyright (C) 2012-2013 B3Partners B.V.
+ * Copyright (C) 2012-2015 B3Partners B.V.
  *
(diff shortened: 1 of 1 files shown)


mprins:
@mtoonen Any thoughts on this?

mtoonen:
Ah, yes. Seems to be a problem introduced when migrating to ExtJS5. The this.config.lox vs. this.lox part is different from those extjs version. Good catch.

PR #502, opened 2015-12-28
pass errorhandler function to ajax request and pass scope to error handler

### LINKED ISSUE #487: Featureinfo control: spinner disappears before all layers have responded
Currently the spinner of the feature info disappears when the first layer responds (irrespective of returning features) 

Consider the following case:
  -  a map with two layers:
     - a sparsely populated layer of buildings (response time ~ 500ms)
     - a layer of parcels (response time ~1500 ms)
  - a feature info control with spinner

When feature info is requested on both layers the spinner starts;
when the [empty] info on the buildings return the spinner is removed but no popup is added; 
("nothing happens" for the user, it seems no info was returned - lets klick again!)
after some time the info on the parcels returns, but the user may have already clicked again.

I can think of several "solutions"

  - wait until the last request has returned (or timed out) before removing the spinner
  - wait for the first response that has a non-empty array of features
  - do nothing
  - educate the user... :bomb: 
  - ...

@geertplaisier We should take a look at this

mprins:
This PR allows passing scope so that the errorhandler function is useful (otherwise the default scope of Window is used) 

As described in https://github.com/flamingo-geocms/flamingo/issues/487#issuecomment-167516763 the errorhandler was not being passed to the Ajax request which would leave the spinner on screen in case of a timeout or error.

close #487

### DIFF
--- a/viewer/src/main/webapp/viewer-html/common/ajax/FeatureInfo.js
+++ b/viewer/src/main/webapp/viewer-html/common/ajax/FeatureInfo.js
@@ -60,8 +60,8 @@ Ext.define("viewer.FeatureInfo", {
                 successFunction(response);
             },
             failure: function(result) {
-                if(failureFunction != undefined) {
-                    failureFunction("Ajax request failed with status " + result.status + " " + result.statusText + ": " + result.responseText);
+                if (failureFunction !== undefined) {
+                    failureFunction("Ajax request failed with status " + result.status + " " + result.statusText + ": " + result.responseText, scope);
                 }
             }
         });
--- a/viewer/src/main/webapp/viewer-html/components/Maptip.js
+++ b/viewer/src/main/webapp/viewer-html/components/Maptip.js
@@ -205,7 +205,7 @@ Ext.define ("viewer.components.Maptip",{
             }
 
             me.onMapData(null, options);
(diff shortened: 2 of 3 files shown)


mprins:
~~the downside of this is that when a maptip is configured the error log fills up with messages "Ajax request failed with status -1 transaction aborted: undefined"~~

handled this in bb3de8189dd43af588a2ce9cfa0b68adaf6c6ebe

viewer/src/main/webapp/viewer-html/common/ajax/FeatureInfo.js
```diff
@@ -60,8 +60,12 @@ Ext.define("viewer.FeatureInfo", {
                 successFunction(response);
             },
             failure: function(result) {
-                if(failureFunction != undefined) {
-                    failureFunction("Ajax request failed with status " + result.status + " " + result.statusText + ": " + result.responseText);
+                if (failureFunction !== undefined) {
+                    // -1 means request aborted; this is not an error, but user initiated
+                    // 0 is a communication error (timeout)
+                    if (result.status > -1) {
```
mprins: @mtoonen Not sure I like this, but it works for the tooltip as that creates a bunch of aborted requests

Issue #504, opened 2015-12-30
database initialization fails


mprins:
When Flamingo creates a new database inserting the metadata fails, the rest works.

logfile:
```
2015-12-22 12:36:48 DEBUG nl.b3p.viewer.solr.SolrInitializer - SolrInitializer initializing
2015-12-22 12:36:48 INFO  nl.b3p.viewer.solr.SolrInitializer - Data dir set D:\Flamingo_data
2015-12-22 12:36:48 DEBUG nl.b3p.viewer.solr.SolrInitializer - Initialize the Solr Server instance
2015-12-22 12:36:49 INFO  net.sourceforge.stripes.vfs.VFS - Using VFS adapter net.sourceforge.stripes.vfs.DefaultVFS
2015-12-22 12:36:49 INFO net.sourceforge.stripes.config.BootstrapPropertyResolver - Class implementing/extending MultipartWrapper found in web.xml: net.sourceforge.stripes.controller.multipart.CommonsMultipartWrapper
2015-12-22 12:36:49 INFO net.sourceforge.stripes.controller.multipart.DefaultMultipartWrapperFactory - Using net.sourceforge.stripes.controller.multipart.CommonsMultipartWrapper as MultipartWrapper implementation.
2015-12-22 12:36:49 INFO  org.stripesstuff.stripersist.Stripersist -
##################################################
# Stripersist Version: 1.0.3, Build: 151:153
##################################################
2015-12-22 12:36:49 INFO  org.stripesstuff.stripersist.Stripersist - Reading persistence.xml from jar:file:/C:/Program%20Files/Apache%20Software%20Foundation/Tomcat%207.0/webapps/viewer/WEB-INF/lib/viewer-config-persistence-4.5.7-SNAPSHOT.jar!/META-INF/persistence.xml.
2015-12-22 12:36:49 INFO net.sourceforge.stripes.config.BootstrapPropertyResolver - Class implementing/extending InitializeSettings found in web.xml: nl.b3p.viewer.config.stripersist.DynamicStripersistInitializer
2015-12-22 12:36:49 INFO nl.b3p.viewer.config.stripersist.DynamicStripersistInitializer - Trying to determine persistence unit from JNDI DataSource database
2015-12-22 12:36:49 INFO nl.b3p.viewer.config.stripersist.DynamicStripersistInitializer - Found persistence unit: viewer-config-postgresql
[...]

PR #505, opened 2015-12-30
Provide a tomcat instance on travis-ci for integration testing

mprins:
This PR introduces a maven profile called "travis-ci" in the viewer artifact that will start a tomcat instance in `pre-integration-test` phase and stop it in `post-integration-test` phase; this leaves the `integration-test` phase to run integration tests.
One integration test was added with two testcases, one to check if flamingo is running (by requesting http://localhost:9090/viewer/) and one testcase to check the flamingo database version. A property file (postgres.properties) is available to modify the database connection params.

The travis configuration has been modified to set up a postgres (9.4) instance with an empty database (flamingo4) with a role (flamingo4/flamingo4).

These changes also make it possible to spin up a local viewer in tomcat on port 9090 using `mvn package tomcat7:run -Ptravis-ci` You may need to change the database params in `viewer/src/test/resources/postgres.properties` to match your environment.
- [x] creates an empty database and user
- [x] spin up a tomcat instance with a viewer webapp loaded
- [x] simple testcase to check if the url is up
- [x] add integration test for database contents

see also #504

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,22 @@ language: java
 
 sudo: false
 
+addons:
+  postgresql: "9.4"
+
+before_install:
+  - psql -U postgres -a -c "CREATE ROLE flamingo4 LOGIN PASSWORD 'flamingo4' SUPERUSER CREATEDB;"
+  - psql -U postgres -a -c 'create database flamingo4;'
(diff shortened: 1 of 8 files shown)


mprins:
@mtoonen This is ready for review

mprins:
added a note to the wiki: https://github.com/flamingo-geocms/flamingo/wiki/Developer-corner

PR #510, opened 2016-01-04
Prevent failure on javadoc errors during release when using java 8 runtime

mprins:
Due to stricter javadoc linting a release build fails because any invalid javadoc will cause an ERROR during the build. see `mvn javadoc:javadoc` output when running on java 8

The quick-and-dirty way is specifying the doclint to none:

``` xml
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-javadoc-plugin</artifactId>
    <configuration>
      <additionalparam>-Xdoclint:none</additionalparam>
    </configuration>
  </plugin>
</plugins>
```

This would need to be set up in a build profile that is automatically activated when building on java 8

The proper way is to fix the javadoc in the artifacts.
- [x] solr-commons
- [x] web-commons
- [x] viewer-commons
- [x] viewer-config-persistence
- [x] viewer
- [x] viewer-admin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,10 @@ install:
 script:
   # execute unit and integration tests
   - mvn -e test verify -B
+  # on java 8 run a javadoc build to check for errors
+  - if [ "$TRAVIS_JDK_VERSION" == oraclejdk8 ]; then
+         mvn javadoc:javadoc;
+    fi
 
 jdk:
   - openjdk7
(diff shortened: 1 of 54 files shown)


mprins:
there are still some warnings regarding [`javax.annotation.security.RolesAllowed`](https://travis-ci.org/flamingo-geocms/flamingo/jobs/100177843#L2765) but no more failures

Issue #507, opened 2015-12-31
Disable autocomplete for username/password fields


matthijsln:
In service details username/password fields are autocompleted with the Flamingo user account details, this is annoying. Rename fields or check if autocomplete="off" works. Also check in other places such as user management

mprins:
I had a try at this a while ago, but the stripes taglib we are using atm turned out not to support this attribute on input or form elements, there is a new release of stripes that may work, but still you may still run into this issue: https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#The_autocomplete_attribute_and_login_fields

matthijsln:
we can use custom attributes with stripes-dynattr taglibs

mprins:
closed with #513

PR #513, opened 2016-01-05
Explicitly set autocomplete to off for username/password fields

mprins:
This PR sets the autocomplete attribute on username/password to off ( #507 )

Even though these are specified I'm still seeing autocomplete behaviour in both Chrome as well as Firefox (as noted in the linked article in the comment: https://github.com/flamingo-geocms/flamingo/issues/507#issuecomment-168408269 )

Also: Chrome fills in the (client-side stored) username to the input closest (in the dom) to the password input (eg the "Telefoon:" input in case of the user editor) if the username input is read-only 
(Firefox will then respect the requested autocomplete)

A hack around his may be to set both username and password to readonly and add a focus handler to remove the readonly attribute, eg:

``` html
<input type="text" readonly  onfocus="this.removeAttribute('readonly');"/>
<input type="password" readonly  onfocus="this.removeAttribute('readonly');"/>
```

see also: https://stackoverflow.com/a/24247840/5454667

Another hack that is reported to work involves adding a dummy, hidden password field to the forms in order to confuse the browser...

### DIFF
--- a/viewer-admin/src/main/webapp/WEB-INF/jsp/security/edituser.jsp
+++ b/viewer-admin/src/main/webapp/WEB-INF/jsp/security/edituser.jsp
@@ -66,11 +66,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                                     </tr>
                                     <tr>
                                         <td>Gebruikersnaam *:</td>
-                                        <td><stripes:text name="username" disabled="${!empty actionBean.user.username}" maxlength="255" size="30"/></td>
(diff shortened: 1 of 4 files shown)


mprins:
IE has totally removed support of autocomplete on password fields: 
- https://lists.w3.org/Archives/Public/public-webapps/2014JanMar/0015.html
- https://security.stackexchange.com/a/104800

mprins:
@flamingo-geocms/committers Anyone have further thoughts about this?

mtoonen:
@geertplaisier You had some ideas how to "fix" this, can you implemented that fix?

PR #521, opened 2016-01-22
Zoom to Feature in attribute list

mprins:
15.0056

Note that feature ids must be consistent across requests; eg. geoserver will "invent" feature ids on a per request basis if no primary key or unique index was discovered in the feature type, this can happen when using views.

### DIFF
--- a/viewer/src/main/webapp/viewer-html/components/AttributeList.js
+++ b/viewer/src/main/webapp/viewer-html/components/AttributeList.js
@@ -151,7 +151,7 @@ Ext.define ("viewer.components.AttributeList",{
                         xtype: 'button',
                         style: { marginLeft: '5px' },
                         itemId: 'zoomToAll',
-                        text: 'Zoom naar alle attributen',
+                            text: 'Zoom naar alle features',
                         disabled: true,
                         scope: this,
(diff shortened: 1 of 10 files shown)


viewer/src/main/webapp/viewer-html/components/AttributeList.js
```diff
@@ -397,8 +409,13 @@ Ext.define ("viewer.components.AttributeList",{
         var attributeList = new Array();
         var columns = new Array();
         var index = 0;
+        var hasGeometry = false;
+        var geometryTypes = ["geometry", "point", "multipoint", "linestring", "multilinestring", "polygon", "multipolygon"];
         for(var i= 0 ; i < attributes.length ;i++){
             var attribute = attributes[i];
+            if(geometryTypes.indexOf(attribute.type) !== -1) {
+                hasGeometry = true;
+            }
```
mprins: probably better to check if the feature type/layer has a `geometryAttribute` or `geometryAttributeIndex` the type is not that interesting. 
Also if the feature does not have a geometry I think the backend will return `success:false`

mprins:
Works just fine for JDBC attribute sources, but WFS sources seem to return an extent as large as the data coverage.. needs work

PR #549, opened 2016-02-26
Filter geometry fields from the featureinfo pop-out window

mprins:
In some cases it's necessary to send the feature geometry to the client when using a feature info, eg. when you want to create a factsheet/printout and want to zoom in to the feature.

The downside is that the second level feature info popup (eg. "toon alle gegevens" link) will show the geometry (or geometries) which is generally unwanted.

This changes the behaviour of the pop-out window to skip rendering geometry fields in the table of all attributes

### DIFF
--- a/viewer/src/main/webapp/viewer-html/common/viewercontroller/ViewerController.js
+++ b/viewer/src/main/webapp/viewer-html/common/viewercontroller/ViewerController.js
@@ -892,6 +892,22 @@ Ext.define("viewer.viewercontroller.ViewerController", {
         return layerObj;
     },
     /**
+     * Get the geometry attributes from the argument.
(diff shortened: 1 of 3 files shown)


mprins:
@geertplaisier @mtoonen this is a fix for the feature info, but not the attributelist; not sure if that is necessary?

I've added a utility function (`getAppLayerGeometryAttributes(appLayer)`) in the ViewerController to get the geometry attributes from an appLayer.

mprins:
rebased and added a flag in the config to disable this behaviour

mtoonen:
I think it would be nice to have it also in the attributelist, so lets make a separate issue for it, so we can merge this.

mtoonen:
See #553

PR #552, opened 2016-03-01
Enhance image output for printing

mprins:
- expose strokeWidth to client
- poor-man's halo on labels of geometries
- use a larger and bold font for the labels


mprins:
@mtoonen Any comments? example print on your desk ;-)

mprins:
Dat is snel... 
I only tested my custom print stylesheet, not the defaults.

Issue #568, opened 2016-03-25
default screen sizes of components 


NiekBouma:
some components have default screen sizes (under water) that are too high. For instance if you forget to set a screen size for the Search component, the search screen in the viewer will be so high, that the information is not visible in the screen.

NiekBouma:
in flamingo4.b3p.nl it is easy to forget to set a screen size. By reusing the defaults again, it is prevented that the information is invisible on the screen.

mtoonen:
Which components?

NiekBouma:
All 45 components that can be selected in the Layout with components. Geert knows more about it.

NiekBouma:
when you enter a value lower than the minimal value for height or width, you do not get an error message and you are able to save the input. However, Flamingo uses the minimal values in the application. This is confusing for users: they should get a red frame around the invalid value(s) and an error message like: 'the value you entered is lower than the minimal value' when trying to save the screen.

geertplaisier:
Added validation message in 30819f66e7c201a4b164a33ee441aa5377d3993b

geertplaisier:
@NiekBouma Please check if this is fixed in the latest snapshot release. If the problem is fixed, please close this issue.

Issue #567, opened 2016-03-25
bookmark field sizes


NiekBouma:
If the bookmark and compact link screens would be higher the complete bookmark could be shown

NiekBouma:
in flamingo4.b3p.nl http://flamingo4.b3p.nl/viewer/app/Solparc%20-%20Meldingen%20openbare%20ruimte/v1 for example, especially when a resolution of 1024 x 768 is used, the bookmark is not completely visible. By making the boxed slightly higher, they would be completely visible.

geertplaisier:
@NiekBouma Please check if this is fixed in the latest snapshot release. If the problem is fixed, please close this issue.

Issue #566, opened 2016-03-25
Scroll bar


NiekBouma:
If a component's screen is too small to show all the elements a scroll bar on the right hand side should emerge to navigate to the elements that are not visible at first.

NiekBouma:
in flamingo4.b3p.nl, if the resolution of the screen is too small in the printing component for instance, a scroll bar on the right hand side should make elements like buttons or text fields visible.

geertplaisier:
@NiekBouma Please check if this is fixed in the latest snapshot release. If the problem is fixed, please close this issue.

NiekBouma:
the problem is fixed. the headers and horizontal scroll bar remain visible at all times.

PR #569, opened 2016-04-02
Impr/configpage refactoring

geertplaisier:
Refactored configPage.jsp / configPage.js and all *-Config.js files. Removed global variables and created configPage component

Also added default values for all popup components (minWidth/minHeight) to fix issues #568, #567 and #566

### DIFF
--- a/viewer-admin/src/main/webapp/WEB-INF/jsp/application/configPage.jsp
+++ b/viewer-admin/src/main/webapp/WEB-INF/jsp/application/configPage.jsp
@@ -29,7 +29,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
     </stripes:layout-component>
 
 
-    <stripes:layout-component name="body">
+    <stripes:layout-component name="body"><%--@elvariable id="actionBean" type="nl.b3p.viewer.admin.stripes.LayoutManagerActionBean"--%>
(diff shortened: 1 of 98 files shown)


mprins:
Just starting an upgrade note here; so for any third-party/custom (java script) components one would need to update the component's config constructor from:
   `constructor: function (parentId, configObject)`
to
   `constructor: function (parentId, configObject, configPage)`
And one should use `this.configObject` class member instead of the injected `config` argument

And add a `getDefaultValues()` function like:

``` javascript
   getDefaultValues: function() {
        return {
            details: {
                minWidth: 400,
                minHeight: 250
            }
        };
```

And there are two new functions:
- `this.getContextpath()`
- `this.getApplicationId()`

The component should be upgraded by adding a `details` to the `config` like:

``` javascript
    config: {

    ....

       details: {
            minWidth: 400,
            minHeight: 250
        }
```

geertplaisier:
Regarding the config constructor you are probably right, the `configPage` variable should be added.

Most components already used `this.configObject` instead of the passed `config` variable so I changed this in a few components for uniformity. Components using `config` directly should still work.

The `getDefaultValues` function is optional, since config objects inherit the function automatically from `ConfigObject.js`

The `getContextpath(), getApplicationId(), getActionBeanUrl(name), getAppConfig()` functions in `ConfigObject.js` are used to get these values passed to `configPage.js` when loading the `configPage.jsp` page

Adding `details: { minWidth: .., minHeight: .. }` is also optional. Is used for popup components which require a minimum width and minimum height to prevent clipping of content.

Issue #576, opened 2016-04-14
Attribute list: shapefile download does not work


NiekBouma:
when you go to flamingo4.b3p.nl, voorzieiningen, Attribute list and you choose 'Shape' and press 'Download' you get the message:
' {
    "message": "Fout bij ophalen features: java.io.IOException: Schema 'wz_voorzieningen' does not exist.",
    "success": false
}'

if you go to Groenbeheer, Attribute list, choose 'BGT' and choose 'Shape' and press 'Download', you get the message:
'{
    "message": "Fout bij ophalen features: org.geotools.data.DataSourceException: Cannot create a shapefile whose geometry type is class com.vividsolutions.jts.geom.Geometry",
    "success": false
}'

Issue #339, opened 2015-04-24
Downloading attributedata gives NullPointerException


geertplaisier:
Probably has something to do with related(join) sources

geertplaisier:
```
2015-04-24 15:55:57 ERROR nl.b3p.viewer.stripes.DownloadFeaturesActionBean - Error loading features
java.lang.NullPointerException
	at nl.b3p.viewer.features.ExcelDownloader.init(ExcelDownloader.java:97)
	at nl.b3p.viewer.stripes.DownloadFeaturesActionBean.convert(DownloadFeaturesActionBean.java:325)
	at nl.b3p.viewer.stripes.DownloadFeaturesActionBean.download(DownloadFeaturesActionBean.java:247)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:483)
	at net.sourceforge.stripes.controller.DispatcherHelper$6.intercept(DispatcherHelper.java:467)
	at net.sourceforge.stripes.controller.ExecutionContext.proceed(ExecutionContext.java:158)
	at net.sourceforge.stripes.controller.BeforeAfterMethodInterceptor.intercept(BeforeAfterMethodInterceptor.java:113)
	at net.sourceforge.stripes.controller.ExecutionContext.proceed(ExecutionContext.java:155)
	at net.sourceforge.stripes.controller.ExecutionContext.wrap(ExecutionContext.java:74)
	at net.sourceforge.stripes.controller.DispatcherHelper.invokeEventHandler(DispatcherHelper.java:465)
	at net.sourceforge.stripes.controller.DispatcherServlet.invokeEventHandler(DispatcherServlet.java:278)
	at net.sourceforge.stripes.controller.DispatcherServlet.service(DispatcherServlet.java:160)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:725)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:301)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:206)
[...]

PR #591, opened 2016-05-03
Fix/attributelistfixes 339 576

mtoonen:


mtoonen:
Fixes #339 and fixes #576

PR #606, opened 2016-06-10
Improved and extended SimpleFilters component

geertplaisier:
- Added "Getalrange" filter to filter between two numbers
- Added possibility to enable checkboxes by default for "Vinkvak" filter
- Improved setting of filter (limited the amount of backend calls)
- Refactored "Vinkvak" and "Keuzerondje" filter config


mprins:
:tada:

PR #611, opened 2016-06-16
Add a css class to labels of the editing fields

mprins:
This adds a css class `editCmpLbl` to the labels so these can be styled using an external stylesheet eg. to change the width of the label

``` css
.editCmpLbl {
    width: 205px !important;
}
.editCmpLbl span {
    width: 200px !important;
}
```

Also use day-month-year formatting for the date picker in the edit fields (and omit the timestamp)


mprins:
see also: 0707ed34802da19ab97d3933d6e2fd944db5a087

PR #614, opened 2016-06-20
Better default application handling

### LINKED ISSUE #597: Error after session timeout on Layoutmanager page
### Where?
viewer-admin, Layoutmanager

### What happened
javax.el.PropertyNotFoundException: Property 'defaultAppId' not found on type nl.b3p.viewer.admin.stripes.LayoutManagerActionBean

### Reproduce
Go to the Layoutmanager page. Wait until your session times out. Refresh the page. Login. Error!

The issue is that the LayoutManagerActionBean forwards to chooseApplication.jsp (r. 188) when there is no active application. But the chooseApplication.jsp expects the property defaultAppId to be present on the bean (which is only defined on the ChooseApplicationActionBean).

mprins:
This adds a  `defaultAppId` property to the LayoutManagerActionBean to prevent a NPE (close #597) and allows setting a `null` value for the default app, effectively unsetting the property.

There is still some strangeness when the session expires; when the session has expired and the user changes the value, the backed sends the login page as a response; this is not handled properly.


mprins:
@geertplaisier can you test?

PR #619, opened 2016-06-27
Upgrades the closure compiler dependency to a current version

mprins:
This was attempted earlier (a4c8f914fc5f5dc80e057be6024a2e47622c5a5a), but reverted as at that time we were still aiming to be java 6 compatible, we're now aiming to be java 7 compatible

see #620

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -146,7 +146,7 @@
             <dependency>
                 <groupId>com.google.javascript</groupId>
                 <artifactId>closure-compiler</artifactId>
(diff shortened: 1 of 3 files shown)


mprins:
I've done some cursory testing, "works for me"

mprins:
this also fixes [CVE-2008-3891](https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2008-3891) which has low to no impact in Flamingo

PR #626, opened 2016-06-30
Don't loop over cookies when there aren't any

### LINKED ISSUE #45: kaartbeeld ontbreekt in factsheet
zowel kaart als overzichtskaart ontbreken in de factsheet

mogelijk relatie met flamingo-geocms/flamingo#617

### LINKED ISSUE #625: No map in PDF of RTF print
### Where?
viewer, printing to PDF or RTF

### What happened
There is no map or overview map on the print (PDF or RTF document)

### What did you expect?
there should be a map in the printout

### Reproduce
try and print a map to pdf or rtf

### Version
4.7.1

this was introduced with #617 

see also B3Partners/flamingo-ibis#45

mprins:
This fixes a NullPointerException in https://github.com/flamingo-geocms/flamingo/blob/master/viewer/src/main/java/nl/b3p/viewer/image/ImageCollector.java#L116 as it's possible that the cookies array is null.

close #625 
close B3Partners/flamingo-ibis#45

### DIFF
--- a/viewer/src/main/java/nl/b3p/viewer/image/ImageCollector.java
+++ b/viewer/src/main/java/nl/b3p/viewer/image/ImageCollector.java
@@ -114,7 +114,8 @@ public class ImageCollector implements Callable<ImageCollector> {
                 String jsessionid = null;
                 String key = "JSESSIONID";
                 if (cookies != null) {
-                    for (Cookie cookie : cookies) {
+                    for (int i = 0; i < cookies.length; i++) {
+                        Cookie cookie = cookies[i];
                         if (cookie.getName().equalsIgnoreCase(key)) {
                             jsessionid = cookie.getValue();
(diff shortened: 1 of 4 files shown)


mprins:
@mtoonen There is something funny with the code or logic; when I run this in the debugger a cookie is set on each request going out at _the time that the preview image is generated_ then after clicking the [printen via PDF|Opslaan als RTF] button we end up in `loadImage()` again but now without any cookies (ofcourse?) because this is a clean request by the server... 

I'm not sure is this is supposed to work this way?..

In my case of B3Partners/flamingo-ibis#45 we only have the second condition, as we never generate a preview... The attached null check fixes the problem, but not sure if it doesn't undo the original intention.

mprins:
oops, I should have pushed 05efdb22b18a58711b84003e9a2705e37562978f to master,not this branch

PR #656, opened 2016-08-05
Pass the layers as well as the store in the ON_LAYERSELECTOR_INITLAYERS

### LINKED ISSUE #51: Bewerken link is alleen zichtbaar in de feature-info als kaartlaag in het startkaartbeeld staat
beschrijving in Mantis: http://mantis.b3p.nl/view.php?id=5960

### LINKED ISSUE #654: edit link of feature info only available for active layers
### Where?
viewer, featureinfo component + edit component

### What happened
the edit link is not available for layers that were not active during control initialization

### What did you expect?
For any editable layer the link should be present if the layer is visible

### Version
4.7.4

mprins:
This will make it possible to get a handle on all layers of the layerselector, eg all eeditable layers instead of just the vidisible editable layers.

This will allow the Edit control to register all editable layers with the FeatureInfo control and not just the visisble editable layers.

close #654
close B3Partners/flamingo-ibis#51

### DIFF
--- a/viewer/src/main/webapp/viewer-html/common/ajax/LayerSelector.js
+++ b/viewer/src/main/webapp/viewer-html/common/ajax/LayerSelector.js
@@ -55,7 +55,7 @@ Ext.define ("viewer.components.LayerSelector",{
         var requestParams = {};
         requestParams[this.config.restriction]= true;
         requestParams["appId"]= appId;
-            var me = this;
+        var me = this;
         if(this.config.layers != null && this.config.layers.length > 0){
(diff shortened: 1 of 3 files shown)


mprins:
close http://mantis.b3p.nl/view.php?id=5960

mtoonen:
Wasn't this event caught by other components as well? Should we change them
too?

mprins:
None of the other components seems to use the passed data, mostly the event is used to set the dropdown to first item.

mtoonen:
Oh, you're right. My bad.

Issue #620, opened 2016-06-27
library updates


mprins:
Flamingo is built using a large number of 3rd party libraries, some of these have not been updated in a long time though the projects maintaining and developing those libraries have release new versions.

A quick list (using `mvn -U org.codehaus.mojo:versions-maven-plugin:2.3:display-dependency-updates > updates.log` command, you can run a quick CVE scan using: `mvn -U org.owasp:dependency-check-maven:1.4.3:check > cve-scan.log`):

  - [x] #619 com.google.javascript:closure-compiler ......... rr2079.1 -> v20160619 
  - [x] #652 com.google.javascript:closure-compiler ......... v20160619 -> v20160713 
  - [x] #632 com.googlecode.json-simple:json-simple ...... 1.1 -> 1.1.1
  - [x] #621 commons-fileupload:commons-fileupload ..... 1.3 -> 1.3.2
  - [x] javax.mail:javax.mail-api ........................................ 1.5.2 -> 1.5.5
  - [x] #667 javax.mail:javax.mail-api ................................1.5.5 -> 1.5.6
  - [x] org.hsqldb:hsqldb ................................................... 2.3.3 -> 2.3.4
  - [x] #623 org.apache.poi:poi ........................................ 3.9 -> 3.14
  - [x] #623 org.apache.poi:poi-ooxml .............................. 3.9 -> 3.14
  - [x] #633 org.quartz-scheduler:quartz ........................... 2.1.5 -> 2.2.3
  - [ ] #595 Solr upgrade
    - org.apache.lucene:lucene-core ..................... 4.6.0 -> 6.1.0
    - org.apache.solr:solr-core ............................... 4.6.0 -> 6.1.0
    - org.apache.solr:solr-solrj ................................ 4.6.0 -> 6.1.0
  - [x] #641 org.json:json ................................... 20090211 -> 20160212
  - [x] #658 org.json:json ................................... 20160212 -> 20160807
  - [x] #668 org.json:json ................................... 20160807 -> 20160810
  - [x] #669 org.mockito:mockito-all .......................... 1.8.4 -> 2.0.2-beta
  - [x] #669 org.powermock:powermock-api-mockito ................... 1.5.5 -> 1.6.5
  - [x] #669 org.powermock:powermock-module-junit4 ................ 1.5.5 -> 1.6.5
  - [x] #661 org.geotools:gt- <all> ......................................... 14.4 -> 14.5
[...]

PR #674, opened 2016-08-22
[WIP] Upgrade b3p-commons-csw 2.0 to 4.2.2

mprins:
upgrade b3p-commons-csw to 4.2.2

see also: #620

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -137,7 +137,7 @@
             <dependency>
                 <groupId>nl.b3p</groupId>
                 <artifactId>b3p-commons-csw</artifactId>
-                <version>2.0</version>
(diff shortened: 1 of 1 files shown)


mprins:
perhaps upgrade to 4.2.3 so the geotools version is in-sync.

see also: 
- B3Partners/b3p-commons-csw#1
- B3Partners/b3p-commons-csw#2

mprins:
This would also drag in some outdated lucene dependencies; they would also need to be upgraded in b3p-commons-csw before this can be merged

mprins:
b3p-commons-csw is not catching up, still at geotools 14, so this is -for the time being- no longer an option

PR #675, opened 2016-08-22
[WIP] improve help access

### LINKED ISSUE #554: better/easier access to help files
see proposal: [improve access to help in viewer admin](https://github.com/flamingo-geocms/flamingo/wiki/improve-access-to-help-in-viewer-admin)

mprins:
This branch will hold the improved help acces as described in https://github.com/flamingo-geocms/flamingo/wiki/improve-access-to-help-in-viewer-admin

@NiekBouma has started copying documentation from [flamingo/viewer-admin/src/main/webapp/resources/html/help.html](https://github.com/flamingo-geocms/flamingo/blob/master/viewer-admin/src/main/webapp/resources/html/help.html) to the wiki

close #554

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,6 @@ Flamingo
 The Repository of Flamingo.
 Check out the [wiki](https://github.com/flamingo-geocms/flamingo/wiki) for more info.
 
-
 [![Travis Build Status](https://travis-ci.org/flamingo-geocms/flamingo.svg?branch=master)](https://travis-ci.org/flamingo-geocms/flamingo)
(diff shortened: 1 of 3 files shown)


mprins:
we may want to think about having a web module that packages all these files into a war that can be installed for those not having access to the (internet) GH wiki and a context config option to point to the correct base url

mprins:
Currently this fails because the Gh wiki pages are served with a restrictive Content Security Policy so most browsers will refuse to load the actual help page with a message like `Refused to display 'https://github.com/flamingo-geocms/flamingo/wiki/viewer_components_Bookmark_Help' in a frame because an ancestor violates the following Content Security Policy directive: "frame-ancestors 'none'".`

mprins:
this will not be happening in the near future

PR #727, opened 2016-12-08
Use tomcat's LockOutRealm to prevent brute-force / dictionary attacks

### LINKED ISSUE #726: Protect against brute force login
### Where?
Login to viewer or admin

### What happened
It is possible to try to login with an infinite number of username/password combinations. This make it possible to do a brute force attack.

### What did you expect?
After a certain number of unsuccessful logins, the user should be excluded from login for a certain time.

### Reproduce
Login with incorrect username/password and repeat.

### Version
any version

### URL
any application

mprins:
Note that, because Flamingo uses form based authentication with the failed login providing the option to retry the HTTP response status is always 200-OK; this is expected behaviour; unless the failed login page page is rendered incorrectly or the failed login page requires authentication (which makes no sense!) 

So the only effect here is actually that the underlying database (JNDI authentication realm) is no longer hit with a request once a user is locked out; an automated brute-force script will still be able to send login requests indefinitely (to prevent this there should be a script that scans catalina.out and the request log an then instructs the firewall to block the misbehaving IP address.

close #726

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,7 +46,7 @@
         <powermock.version>1.6.6</powermock.version>
         <hsqldb.version>2.3.4</hsqldb.version>
         <test.persistence.unit>viewer-config-hsqldb</test.persistence.unit>
-        <postgresql.version>9.4.1209.jre7</postgresql.version>
+        <postgresql.version>9.4.1212.jre7</postgresql.version>
         <apache.poi.version>3.15</apache.poi.version>
         <!-- skip integration tests by default, override this property from commandline
         or as part of a profile, see eg. 'travis-ci' profile in viewer -->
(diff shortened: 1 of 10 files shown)


mprins:
Currently the actual testcase for the viewer-admin is ignored because there is a problem setting up the jndi authentication realm from within the tomcat7-maven-plugin that I can't figure out.

mtoonen (commented):
Wow, this looks REALLY good!!1!

mtoonen (changes requested):
Make the jndi work for the authentication realm, minion!

mtoonen (approved):

mtoonen:
It's amazing, Mark!

PR #731, opened 2016-12-09
Add a bean + client error handler to enable clients to send messages to the application log

mprins:
Will log any message to the application log, if that is set up for ~~WARN~~ DEBUG severity logging (by default it isn't);

  - [x] Maybe we want this to only log things when logging severity is DEBUG (off by default) as this makes it possible for clients to write (to the log file) on the server., 
 - [x] Maybe we want to add application info (id/name/version) as well?


mtoonen (changes requested):
* Maybe separate the jsp and js code
* Separate logfile for clientside

mprins:
I think it's better to have an integral logfile so you don't spend a lot of time cross-referencing the logs and you don't spend time on telling a sysadmin on the other end of the line to also send this other file

mprins:
We may want to revisit this for further enhancements

PR #734, opened 2016-12-19
Use the cyclomedia service directly

mtoonen:


viewer/src/main/webapp/viewer-html/components/Cyclorama-config.js
```diff
@@ -173,6 +173,7 @@ Ext.define("viewer.components.CustomConfiguration",{
                 store.add({id: id, serviceId: appLayer.serviceId, layerName: appLayer.layerName, alias: appLayer.alias});
             }
         });
+        store.add({id: "-666", serviceId: "-666", layerName: "Cyclomedia direct", alias: "Cyclomedia direct"});
```
mprins, line 176: :smiling_imp:

PR #736, opened 2016-12-22
Feat/ip access

mtoonen:


viewer/src/main/java/nl/b3p/viewer/util/IPAuthenticationFilter.java
```diff
@@ -0,0 +1,347 @@
...
+            FilterChain chain)
+            throws IOException, ServletException {
+        
+        log.debug("IPAuthenticationFilter:doFilter()");
+        HttpServletRequest request = (HttpServletRequest) r;
+        HttpSession session = request.getSession();
+        User u = null;
+        if(session.getAttribute(IP_CHECK) == null  && session.getAttribute(USER_CHECK) == null){
+            
+            String ipAddress = request.getHeader("X-FORWARDED-FOR");
```
mprins: You're using both `X-FORWARDED-FOR` `X-Forwarded-For` I would stick to the latter (mixed-case)

mprins: also, apparently, the forwarded-for can hold multiple adresses [1] so you may want to get the substring up to the first comma.


[1] https://nl.wikipedia.org/wiki/X-Forwarded-For

viewer/src/main/java/nl/b3p/viewer/util/IPAuthenticationFilter.java
```diff
@@ -0,0 +1,347 @@
...
+            
+            if(possibleUsers.isEmpty()){
+                log.debug("No possible users found for ip");
+            }else if( possibleUsers.size() == 1){
+                u = possibleUsers.get(0);
+                Hibernate.initialize(u.getGroups());
+                session.setAttribute(IP_CHECK, ipAddress);
+                session.setAttribute(USER_CHECK, u);
+            }else{
+                log.debug("Too much possible users found for ip.");
```
mprins: Too many... also I would show this as a warning or error as it is a configuration problem

viewer-admin/src/main/webapp/resources/js/security/ipmanager.js
```diff
@@ -0,0 +1,77 @@
...
+        });
+    },
+
+    addIpField: function(val) {
+        var idx = this.formPanel.items.length - 1;
+        this.formPanel.insert(idx, {
+            xtype: 'textfield',
+            name: "ips",
+            value: val || '0.0.0.0',
+            regex: /[0-2][0-9]{0,2}\.[0-2][0-9]{0,2}\.[0-2][0-9]{0,2}\.[0-2][0-9]{0,2}/
```
mprins: this seems to exclude entering an address with  wildcard, but there is Java code using that... 

Also you may want to strip leading/trailing spaces here or when storing in the database

mprins (changes requested):
I think I found some minor issues that should not be too hard to address

PR #754, opened 2017-01-23
Fixed database: removed faulty toplayers, changed constraint on oracl…

mtoonen:
…e to be on correct table


mprins (approved):
all tests now passing

PR #782, opened 2017-02-21
Add event-listening in vector layer for undo/redo/cancel

mprins:
This adds an event listener when the draw control is activated so drawing can be un-done using `CTRL-Z`, re-done using `CTRL-Y` and cancelled using `ESCAPE` keys. The eventlistener is removed when closing the control.

This affects any control drawing on the OpenlayersVectorLayer (Drawing, Edit, Split) but not he Modify action(as that does not draw). For sanity sake the circle tool is excluded.

### DIFF
--- a/viewer/src/main/webapp/viewer-html/common/viewercontroller/controller/VectorLayer.js
+++ b/viewer/src/main/webapp/viewer-html/common/viewercontroller/controller/VectorLayer.js
@@ -16,8 +16,9 @@
  */
 
 /**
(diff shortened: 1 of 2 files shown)


mtoonen:
Can't this be done in the superclass? That way OpenLayers 3/4 can profit as
well, without much duplicate code.

mprins:
Think I nailed it now.
atm. OL4 doesn't have undo/redo functions on the draw interaction afaics.

PR #786, opened 2017-02-28
Fixed #240: you can add a level with layers in it's sublevels

### LINKED ISSUE #240: Adding a level with levels to Startkaartbeeld only possible with a layer in parent
Example
-Level 1
-- Level 2
---Layer A
-- Level 3
--- Layer B

It is not possible to add Level 1 to the startkaarbeeld, but it is when you add a layer to Level 1 (which you could remove afterwards as a workaround)

mprins:
@mtoonen It seems fix this in a branch that was already merged+deleted on GH and then pushed the branch again, so now it's dangling.. or maybe not.

You should probably just merge this PR and delete the branch again I guess


mtoonen:
Thanks. Commited this fix to the wrong branch, but cherry picked it into master, so all is good :-)

PR #796, opened 2017-03-30
Link to AttributeList in featureinfo doesn't register to all layers (only the startuplayers)

geertplaisier:


mtoonen (commented):
Does it work when new layers are added (via selectionmodule), or layers which were not visible at startup, were being switched on?

mtoonen (approved):

PR #810, opened 2017-04-10
Add an option to provide attributes without alias substitution

mprins:
This enables matching the attributes on the field name; as eg for showing a previous version of attributes in the editor.

Previous Flamingo behaviour is preserved by defaulting to `true`

see B3Partners/flamingo-ibis#68


mtoonen (approved):
Looks fine

PR #815, opened 2017-04-10
Fixes #812: allow data from current request

### LINKED ISSUE #812: feature info randomly fails to render
### Version
Which version of flamingo do you use: 4.8.0 / ibis 2.14

### Browser info
Which browser are you using? FF, IE, Chrome

### URL
http://bedrijventerreinen.gelderland.nl/app/IBISPROVINCIE (for credentials contact @mprins)

### Where?
Viewer, FeatureInfo component

### What happened
FeatureInfo fails to render attributes of of Kavels layer occasionally (variably 2/10 to 7/10 times) when several other layers are enabled (Bedrijventerreinen, uitgifte historie, vestigingen van bedrijven).
No errors are logged and the XHR for alle layers are executed successfully and return data

### What did you expect?
All FeatureInfo should be rendererd

### Reproduce
How can we reproduce this? 

  1. login to the application
  2. switch on layers "Uitgifte historie" and "Vestigingen van bedrijven"
  3. zoom in to see  "Vestigingen van bedrijven"
  4. get featureinfo in the area of a Bedrijventerrein
  5. observe seemingly random failure

mtoonen:


mprins (approved):
kudos! :1st_place_medal:

PR #827, opened 2017-04-20
Fixed (?) retrieving the geometry from the attributes. Closes #826. I…

### LINKED ISSUE #826: failure to parse wkt from json attributes response in snapping component
### Version
Which version of flamingo do you use: 4.8.1-rc1

### Browser info
any

### URL
Where can we find an application which has the bug?

### Where?
snapping

### What happened
failure to parse wkt from json attributes response around line 226 in the snapping component triggering a NPE in the openlayers code.
snapping layer has joins and relates and it seems the geometryAttributeIndex is "off" so openlayers does not get wkt but something else..

### What did you expect?
should work as in 4.8.0

mtoonen:


mprins:
I cannot remember why that code-convolution is there except the comment:
> test for keys in element, some WFS just return all attributes anyway..

maybe this behaviour changed when the geotools WFS stack was changed out

mprins (approved):
can't remember why I didn't use named object members instead of property indexes... but since we're only asking for a single attribute it should always be `c0`

PR #856, opened 2017-05-22
Adding codecoverage hooks

mtoonen:


.travis.yml
```diff
@@ -33,6 +33,8 @@ script:
   - mvn -e clean verify -B -Ptravis-ci -pl '!viewer-admin'
   # run integration tests on viewer-admin module only
   - mvn -e clean verify -B -Ptravis-ci -pl 'viewer-admin'
+  # run code coverage
+  - mvn cobertura:cobertura
```
mprins, line 37: this will skip any integration tests from being reported. probably `cobertura:cobertura-integration-test`should be used  see: http://www.mojohaus.org/cobertura-maven-plugin/cobertura-integration-test-mojo.html

I'll set up a PR to see if that works

PR #858, opened 2017-05-22
Better test coverage reporting

mprins:
Attempt to fix integration test coverage report and start using jacoco instead of cobertura

This should give us at least unit test coverage, integration test coverage is painful to set up because we run inside tomcat which is a forked process inside maven and it requires trickery to get the javaagent into tomcat.
See eg. https://stackoverflow.com/questions/16018698/tomcat7-maven-plugin-and-jacoco (and maybe links below) for some options and workarounds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,8 @@ script:
   # run integration tests on viewer-admin module only
   - mvn -e clean verify -B -Ptravis-ci -pl 'viewer-admin'
   # run code coverage
-  - mvn cobertura:cobertura -Ptravis-ci
+  #- mvn cobertura:cobertura -Ptravis-ci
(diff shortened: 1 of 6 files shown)


mtoonen (approved):

mprins:
perhaps we would be better off using jacoco, some links:
- https://docs.codecov.io/docs/supported-languages
- https://github.com/codecov/example-java-maven
- https://github.com/codecov/example-java
- https://github.com/codecov/example-java/blob/master/pom.xml#L38-L56
- https://www.petrikainulainen.net/programming/maven/creating-code-coverage-reports-for-unit-and-integration-tests-with-the-jacoco-maven-plugin/
- https://stackoverflow.com/questions/37485133/have-integration-tests-in-jacoco-it-directory

mprins:
image

PR #888, opened 2017-08-17
Feat/spatial filter based on other layer

mtoonen:
makes it possible to do cross-layer filtering

### DIFF
--- a/viewer/src/main/java/nl/b3p/viewer/util/FlamingoCQL.java
+++ b/viewer/src/main/java/nl/b3p/viewer/util/FlamingoCQL.java
@@ -20,13 +20,13 @@ import com.vividsolutions.jts.geom.Geometry;
 import com.vividsolutions.jts.geom.GeometryCollection;
 import com.vividsolutions.jts.geom.GeometryFactory;
 import com.vividsolutions.jts.geom.PrecisionModel;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 import javax.persistence.EntityManager;
 import nl.b3p.viewer.config.app.ApplicationLayer;
 import nl.b3p.viewer.config.services.GeoService;
 import nl.b3p.viewer.config.services.Layer;
 import static nl.b3p.viewer.util.FeatureToJson.MAX_FEATURES;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.geotools.data.FeatureSource;
 import org.geotools.data.Query;
 import org.geotools.feature.FeatureIterator;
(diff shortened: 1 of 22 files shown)


viewer/src/main/java/nl/b3p/viewer/util/FlamingoCQL.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    public static String processFilter(String filter, EntityManager em) throws CQLException {
+        if (filter.contains(BEGIN_PART)) {
+            filter = replaceApplayerFilter(filter, em);
+        }
+        return filter;
+    }
+
+    protected static String replaceApplayerFilter(String filter, EntityManager em) throws CQLException {
```
mprins: why is this "protected"?

viewer/src/main/java/nl/b3p/viewer/util/FlamingoCQL.java
```diff
@@ -0,0 +1,172 @@
...
+        // Rewrite APPLAYER filter to GEOMETRY filter, so it can be used for filtering other features
+        String geometryFilter = rewriteAppLayerFilter(appLayerPart, em);
+
+        String beginpart = filter.substring(0, begin);
+        String endpart = filter.substring(endIndex + 1);
+        String result = beginpart + geometryFilter + endpart;
+        return result;
+    }
+
+    protected static String rewriteAppLayerFilter(String applayerfilter, EntityManager em) throws CQLException {
```
mprins: why is this "protected"?

viewer/src/main/java/nl/b3p/viewer/util/FlamingoCQL.java
```diff
@@ -0,0 +1,172 @@
...
+            try {
+                Geometry gc = new GeometryCollection(null, gf);
+                while (it.hasNext()) {
+                    SimpleFeature f = it.next();
+                    Geometry g = (Geometry) f.getDefaultGeometry();
+                    if (g != null) {
+                        gc = gc.union(g);
+                    }
+                }
+
```
mprins, line 160: I would add a self-union here (`gc.union().toText();`) to clean up the geometry

viewer/src/main/java/nl/b3p/viewer/util/FlamingoCQL.java
```diff
@@ -0,0 +1,172 @@
...
+            Query q = new Query(fs.getName().toString());
+            if (filter != null && !filter.isEmpty()) {
+                Filter attributeFilter = ECQL.toFilter(filter);
+                attributeFilter = (Filter) attributeFilter.accept(new ChangeMatchCase(false), null);
+
+                q.setFilter(attributeFilter);
+            }
+
+            q.setMaxFeatures(MAX_FEATURES);
```
mprins, line 147: not sure how big a problem this would be, but having a max amount of features could lead to unexpected results if there are many more features.
Ideally a count should be done first to see how many features would fit the filter and if necessary several requests made; though on many a WFS the count will probably not give a true answer.

viewer/src/test/java/nl/b3p/viewer/util/FlamingoCQLTest.java
```diff
@@ -0,0 +1,121 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
mprins: remove this and add copyright

mprins (changes requested):
Nice! I have some small remarks and questions

mprins (approved):

mprins:
I noticed surefire crash with "Process Exit Code: 137" which basically means maven was killed externally using SIGKILL signal. We may want to check the -X options in surefire plugin and or limit the number of running threads using -T1

PR #889, opened 2017-08-22
use a native query to determine `isMashup` and cache the value

mprins:
Also use explicit lazy loading for application details.

it turns out that -on oracle- Application#isMashup() is slow because it needs to handle a potentially large amount of data in the application_details value CLOBs -specifically cachedExpandedSelectedContent and cachedSelectedContent can be in the order of 1.5-2 MB- 
(postgresql is much better at lazy loading)

reference [mantis 6724](http://mantis.b3p.nl/view.php?id=6724)

### DIFF
--- a/viewer-admin/src/main/java/nl/b3p/viewer/admin/stripes/ChooseApplicationActionBean.java
+++ b/viewer-admin/src/main/java/nl/b3p/viewer/admin/stripes/ChooseApplicationActionBean.java
@@ -216,7 +216,7 @@ public class ChooseApplicationActionBean extends ApplicationActionBean {
         }
         return new ForwardResolution(EDITJSP);
     }
-    
+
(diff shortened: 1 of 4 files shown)


mprins:
some stats for ...viewer-admin/action/chooseapplication/getGridData (basically clicking on the "Kies een Applicatie" link

Against an Oracle 12.1. DB running on virtualbox VM with 3GB ram and 3 processors

| times in sec   | 1  |2   |3   |4   |
|---:|---:|---:|---:|---:|
| before  | 10.98| 10.24| 9.95| 10.02|
|after      | 1.37| 0.81| 0.72| 0.59 |

mprins:
PostgreSQL lijkt ook (nog) prima te werken :8ball:

mtoonen (approved):

PR #907, opened 2017-10-12
introduce OpenLayers 4 viewer

mprins:
This would bring in Martijn's work and make it possible to define viewers based on OpenLayers 4


mprins:
some TODO's while I scanned through the PR:
- [x] remove `viewer-admin/hs_err_pid3425.log`
- [x] load latest OL4 libs
- [x] rename references to `openlayers3` to `openlayers4`
- [x] ~~think about having the whole OL4 source tree or just the compiled (ol.js and ol-debug.js) in our source tree (I think we don't have the infrastructure in place to compile anyway)~~ I have prepared #912 to remove the OL4 sources

mprins:
I created an OL4 viewer with legend, toc and scalebar (no tools) and am seeing the following stacktrace when loading the viewer:
```
Uncaught TypeError: Cannot read property 'on' of null
    at constructor (OpenLayers4DefaultTool.js:52)
    at new constructor (ext-all-debug.js:7660)
    at constructor.checkTools (OpenLayers4MapComponent.js:84)
    at OpenLayers4MapComponent.js:65
```
it is then not possible to drag/pan, drag/zoom, scroll-zoom works.

it seems that `this.getFrameworkTool()` in OpenLayers4DefaultTool returns `null` as it is being added by `checkTools` in OpenLayers4MapComponent

adding some tools (identify,pan,zoo in/out) resolves the issue

mprins:
OpenLayers 5 having been release I don't think we should consider OL4 any more

mprins:
Closing in favour of #1386

PR #912, opened 2017-10-20
[wip] remove Openlayers 4 source tree

mprins:
after merging #907 this branch should be rebased and could then be merged to remove the Openlayers 4 source tree


mprins:
OpenLayers 5 having been release I don't think we should consider OL4 any more

PR #970, opened 2018-01-12
impr/maintain attribute order

geertplaisier:
Use JSONArray instead of JSONObject for attributes to maintain order of the attributes.
Currently used by FeatureInfoReport & FeatureInfo component

### DIFF
--- a/viewer/src/main/java/nl/b3p/viewer/stripes/FeatureInfoActionBean.java
+++ b/viewer/src/main/java/nl/b3p/viewer/stripes/FeatureInfoActionBean.java
@@ -100,6 +100,9 @@ public class FeatureInfoActionBean implements ActionBean {
     @Validate
     private boolean graph = false;
 
+    @Validate
+    private boolean ordered = false;
+
     private Layer layer;
 
     //<editor-fold defaultstate="collapsed" desc="getters and setters">
@@ -197,6 +200,14 @@ public class FeatureInfoActionBean implements ActionBean {
 
     public void setGraph(boolean graph) {
         this.graph = graph;
+    }
+
+    public boolean isOrdered() {
+        return ordered;
+    }
+
+    public void setOrdered(boolean ordered) {
+        this.ordered = ordered;
     }
 
     public Layer getLayer() {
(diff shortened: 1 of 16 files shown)


viewer/src/main/java/nl/b3p/viewer/util/FeaturePropertiesArrayHelper.java
```diff
@@ -0,0 +1,46 @@
+package nl.b3p.viewer.util;
```
mprins, line 17: this should probably have a gpl header

viewer/src/main/java/nl/b3p/viewer/util/FeatureToJson.java
```diff
@@ -303,12 +349,24 @@ private JSONObject populateWithRelatedFeatures(JSONObject j,SimpleFeature featur
...
+                    relatedFeatures.put("related_featuretypes", related_featuretypes);
+                    ordered.put(relatedFeatures);
+                }
             }
         }
         return j;
     }
 
+    private void concatArray(JSONArray arr1, JSONArray arr2) throws JSONException {
+        for (int i = 0; i < arr2.length(); i++) {
```
mprins, line 364: probably there is a JSONArray#addAll() because I think there is a java collection backing the JSONArray, maybe depends which impl this is... see eg. https://docs.oracle.com/javaee/7/api/javax/json/JsonArray.html

mprins (approved):

viewer/src/main/java/nl/b3p/viewer/util/FeatureToJson.java
geertplaisier: It uses an other implementation: http://www.docjar.com/docs/api/org/json/JSONArray.html which does not have an addAll/putAll method

mtoonen (approved):

viewer/src/main/java/nl/b3p/viewer/stripes/FeatureInfoActionBean.java
```diff
@@ -360,7 +371,7 @@ public Resolution info() throws JSONException {
     protected JSONArray executeQuery(ApplicationLayer al, SimpleFeatureType ft, FeatureSource fs, Query q)
             throws IOException, JSONException, Exception {
 
-        FeatureToJson ftjson = new FeatureToJson(arrays, edit, graph, attributesToInclude);
+        FeatureToJson ftjson = new FeatureToJson(arrays, edit, graph, false, false, attributesToInclude, ordered);
```
mprins, line 374: the 4th argument should be `true`

PR #972, opened 2018-01-15
update flamingo.solr.schedule docs

mprins:
update `flamingo.solr.schedule` docs, the code expects a cron-expression not an interval in minutes


mprins:
see also #973

mtoonen (approved):

PR #983, opened 2018-01-29
Enhance scale calculus for zoomToScale

mprins:
Tweak ViewerController#calculateScaleCorrection() to multiply with an extra factor depending on the type of wms to better match resolutions with max/min scale denominators in sld, magic numbers...

fixes [mantis 6914](http://mantis.b3p.nl/view.php?id=6914)


mtoonen (commented):
Looks good!

mtoonen (approved):

PR #984, opened 2018-01-30
reduce feature report geometry size if necessary

mprins:
This will reduce the feature report geometry in size by simplifying, starting off with 1 (meter) and  each iteration multiply with 10, max 4 steps, so 1,10, 100, 1000 if geom still to large bail out and use bbox.
There will be a note on the report.

the limit is 2MB-100KB


KJLammers:
[Kaart_30-01-2018 (4).pdf](https://github.com/flamingo-geocms/flamingo/files/1677122/Kaart_30-01-2018.4.pdf)
[Kaart_30-01-2018 (5).pdf](https://github.com/flamingo-geocms/flamingo/files/1677123/Kaart_30-01-2018.5.pdf)

mtoonen (approved):

PR #985, opened 2018-02-08
WIP start testing with ibmjava8

mprins:


mprins:
travis fails to understand ibm java 8, giving up

PR #986, opened 2018-02-08
Bugs/svg and featureinfo issues

geertplaisier:
Fixes 3 issues:
- Clicking tool buttons was difficult in IE with SVG icons enabled. You had to click within the button area but outside the icon itself
- When SVG was enabled and there where multiple tools configured and also Next/Previous extent tools, the other tools would not show up in IE
- In FeatureInfo/Maptip, when all layers are off when starting the viewer a spinner would occur when clicking the map but the spinner would never be removed because no request was done.


mprins (approved):
not tested the SVG fix, as I don't have IE at hand, featureinfo looks good to me

PR #1011, opened 2018-03-29
Feat/multiple attribute source searching

mtoonen:


mprins (approved):
Looked at the code, I don't have a solr configured so not tested, but already in a release candidate anyway :1st_place_medal:

PR #1012, opened 2018-03-29
Give the connection scheme to the proxy http client

mprins:
Give the connection scheme to the proxy http client to make preemptive authentication work for https sites.

fixes [mantis 10740](http://mantis.b3p.nl/view.php?id=10740)


mprins (commented):
XP team effort :smile_cat:

mtoonen (approved):

PR #1016, opened 2018-04-05
Drawing: Remove selected object btn. visible again

geertplaisier:
Button was hidden with default width/height settings. Improved layout + made button visible again


mprins (approved):
LGTM

PR #1021, opened 2018-04-20
Use flamingocql for translating cql strings to geotools Filters

mtoonen:


mtoonen:
Fixes mantis issue http://mantis.b3p.nl/view.php?id=10864

mprins (approved):

PR #1022, opened 2018-04-25
Feat/datepicker simple filter

mtoonen:


viewer-config-persistence/src/main/java/nl/b3p/viewer/config/services/FeatureSource.java
```diff
@@ -130,15 +126,19 @@ public String getProtocol() {
...
             FilterFactory2 ff = CommonFactoryFinder.getFilterFactory2(null);
             Function unique = ff.function("Collection_Unique", ff.property(attributeName));
             Filter notNull = ff.not( ff.isNull( ff.property(attributeName) ));
-            
-            org.geotools.data.Query q = new org.geotools.data.Query(sft.getTypeName(), notNull);
+            Filter f = notNull;
+            if(filter != null){
+                f = ff.and(notNull, filter);
+            }
+            org.geotools.data.Query q = new org.geotools.data.Query(sft.getTypeName(), f);
```
mprins, line 139: I think this should use an import statement for Query, also on line 171

mprins (changes requested):
I don't like the reformatting of the context file

viewer-config-persistence/src/main/java/nl/b3p/viewer/config/services/FeatureSource.java
mtoonen: Necessary because of import java.persistence.* (which also has a Query class)

mprins (approved):

PR #1023, opened 2018-04-25
Fixes mantis issue 10289: mashup should honor rights of motherapplication

mtoonen:
update authorizationsmodified of mashups (and set cache to dirty) after motherapplication changed settings
http://mantis.b3p.nl/view.php?id=10289


mprins (approved):
makes sense

PR #1029, opened 2018-05-15
Bug/fix click pan

mtoonen:
Fixes mantis 10471.
I think this is because of https://gis.stackexchange.com/questions/20859/will-mousedown-events-fire-with-touch-events-on-openlayers-mobile: click event is never thrown when a clickhandler is registered. Mouseup and mousedown are, but with a drag this never results in a click. Adding a vectorlayer fixes this. I know it's ugly, but after 8/10 hours, I give up.


mvdstruijk (commented):
This was also the only solution that i could find. Adding a vector layer seems to register the browser events again.

Maybe change the pixelTolerance to a variable so it can bet set in the viewer-admin?

mvdstruijk (approved):
pixelTolerance for now is static set on 10.

PR #1030, opened 2018-05-17
Feat/print attributes

mtoonen:


viewer/src/main/webapp/viewer-html/components/Print.js
```diff
@@ -1019,6 +1024,17 @@ Ext.define ("viewer.components.Print",{
                 config.overview.protocol = url.toLowerCase().indexOf("getmap") > 0 ? 'WMS' : 'IMAGE';
             }
         }
+        if(config.includeAttributes){
+            var attributeList = this.viewerController.getComponentsByClassName("viewer.components.AttributeList")[0];
+            var appLayer = attributeList.layerSelector.getValue();
```
geertplaisier: We should add a check here to check the existence of the AttributeList component. If this component is not configured you will get an error on this line.

viewer/src/main/webapp/viewer-html/components/Print.js
```diff
@@ -1019,6 +1024,17 @@ Ext.define ("viewer.components.Print",{
...
+            var appLayerId = appLayer.id;
```
geertplaisier: Add a check if appLayer is not null here, otherwise you would get an error on this line. This might be the case if non of the configured layers for the AttributeList component are enabled.

geertplaisier (changes requested):

geertplaisier (approved):

PR #1049, opened 2018-05-31
Fixes toolclicks on Android

geertplaisier:
Updated buttonclick.js by latest version: https://github.com/openlayers/ol2/blob/master/lib/OpenLayers/Events/buttonclick.js


mtoonen (approved):

mtoonen:
Hmm, this fix is not included in the minified openlayers, so it will only work with ?ol=debug?

mprins (approved):
LGTM

PR #1085, opened 2018-06-12
Fixes #761: Set the default srs on the query. For now RD.

### LINKED ISSUE #761: Pass projection/srsname to wfs featureinfo request
Currently, when a featureinfo request is done, the point doesn't have the srsname of the coordinates, even though it's a valid attribute of a gml:point.
<SERVER>&VERSION=1.0.0&MAXFEATURES=11&FILTER=<Filter+xmlns="http://www.opengis.net/ogc"+xmlns:gml="http://www.opengis.net/gml"><DWithin><PropertyName>geom</PropertyName><gml:Point><gml:coordinates+decimal="."+cs=","+ts="+">173280.0,557853.44</gml:coordinates></gml:Point><Distance>0.0</Distance></DWithin></Filter>&TYPENAME=test:gem_2014_new

mtoonen:


mprins (approved):
not happy with yet another place where we hardcode srs, but since we are still looking for funding for multi-srs support I guess it's OK

PR #1101, opened 2018-06-21
First working version of the external link opener

mtoonen:
New component. Provides the ability to create links with some mapvariables. Links are accessible via the right mouse button.


mprins (changes requested):
It seems to be missing the svg icons, also I don't think the context.xml changes are more than whitespace?...

viewer/src/main/webapp/viewer-html/components/ExternalLinks.js
```diff
@@ -0,0 +1,123 @@
...
+        var map = this.config.viewerController.mapComponent.getMap();
+        var coord = map.pixelToCoordinate(x, y);
+        var extent = map.getExtent();
+        var dest = new Proj4js.Proj("EPSG:4236");
+        var source = new Proj4js.Proj("EPSG:28992");
+        var wgs = new Proj4js.Point(coord.x,coord.y);
+        Proj4js.transform(source,dest,wgs);
+        
+        var currentVariables = {
+            X: coord.x,
```
mprins, line 105: we may want to round these values to realistic values, eg. 1/10th meter and or 1/1000th degrees

mtoonen:
No svg's were used, context.xml didn't change (only a bit more readable)

mprins (approved):

PR #1102, opened 2018-06-21
First working version of DownloadWKT component

mtoonen:
New component. User can draw a polygon/box, with some extra fields, which are then written to the server.


viewer/src/main/webapp/viewer-html/components/DownloadWKT-config.js
```diff
@@ -0,0 +1,129 @@
...
+            for (var i = 0; i < types.length; i++) {
+                this.createRow(types[i]);
+            }
+        }
+    },
+    addForm: function() {
+        this.form.add({ 
+                xtype: 'textfield',
+                fieldLabel: 'Basispad *',
+                name: 'basePath',
```
mprins, line 37: I think this should default to the flamingo data dir or the java.io.tmpdir, perhaps it it should not even be configurable from the viewer-admin at all.

viewer/src/main/webapp/viewer-html/components/DownloadWKT.js
```diff
@@ -0,0 +1,204 @@
...
+
+Ext.define("viewer.components.DownloadWKT", {
+    extend: "viewer.components.Component",
+    vectorLayer: null,
+    config: {
+        types: null
+    },
+    /**
+     * @constructor
+     * creating a print module.
```
mprins: ??

viewer/src/main/webapp/viewer-html/components/DownloadWKT-config.js
mtoonen: File must be saved to disk, so an automated process can pick it up for processing. Asked if it was possible for that folder to be in flamingo_data_dir.

mprins (approved):

PR #1105, opened 2018-06-25
Fix for wrong columns in fileupload

mtoonen:


mprins (changes requested):
I think we need to add an `alter table ...` to migrate as well because this upload has already been released (5.1.5 and v5.2.0-rc1)

5.3.0 was released today as well

mprins:
I've updated this to drop the `file_upload` table if it's there and recreating with a higher database version, also rebased to current master. The previous upgrade was replaced with a dummy-operation.
This should take care of upgrading early adopters using 5.1.5, 5.3.0 (or 5.2.0-RC1 which would require a backport)

@mtoonen WDYT?

mprins (approved):

Issue #1115, opened 2018-06-27
Refresh of (extended) feature info needed after edit


cvanlith:
### Version
5.1.5

### Browser info
chrome 67

### URL
flamingo4.b3p.nl mobiel

### Where and/or What
extended info with edit link

### What happened
after edit the info visible in the extended info panel is not refreshed

### What did you expect?
after edit the (extended) feature info panel should be refreshed

### Reproduce
  1. have info in extended feature panel
  2. press edit
  3. edit one of the values and save
  4. info in extended feature info panel remains the same.

mprins:
see #1121

PR #1121, opened 2018-06-29
Refresh feature info after editing feature

geertplaisier:
Solution for  #1115

### DIFF
--- a/viewer/src/main/webapp/viewer-html/components/ExtendedFeatureInfo.js
+++ b/viewer/src/main/webapp/viewer-html/components/ExtendedFeatureInfo.js
@@ -73,7 +73,8 @@ Ext.define ("viewer.components.ExtendedFeatureInfo",{
         this.getViewerController().mapComponent.getMap().addListener(viewer.viewercontroller.controller.Event.ON_GET_FEATURE_INFO,this.onFeatureInfoStart,this);
         //listen to a extent change
         this.getViewerController().mapComponent.getMap().addListener(viewer.viewercontroller.controller.Event.ON_CHANGE_EXTENT, this.onChangeExtent,this);
-        this.config.viewerController.addListener(viewer.viewercontroller.controller.Event.ON_EDIT_SUCCESS, this.featureEdited, this);
+        this.getViewerController().addListener(viewer.viewercontroller.controller.Event.ON_EDIT_SUCCESS, this.featureEdited, this);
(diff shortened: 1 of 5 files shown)


viewer/src/main/webapp/viewer-html/components/ExtendedFeatureInfo.js
```diff
@@ -73,6 +73,7 @@ Ext.define ("viewer.components.ExtendedFeatureInfo",{
         this.getViewerController().mapComponent.getMap().addListener(viewer.viewercontroller.controller.Event.ON_GET_FEATURE_INFO,this.onFeatureInfoStart,this);
         //listen to a extent change
         this.getViewerController().mapComponent.getMap().addListener(viewer.viewercontroller.controller.Event.ON_CHANGE_EXTENT, this.onChangeExtent,this);
+        this.config.viewerController.addListener(viewer.viewercontroller.controller.Event.ON_EDIT_SUCCESS, this.featureEdited, this);
```
mprins: odd to use `this.config.viewerController` here while bove we use `this.getViewerController()`

viewer/src/main/webapp/viewer-html/components/FeatureInfo.js
```diff
@@ -66,7 +67,8 @@ Ext.define ("viewer.components.FeatureInfo",{
         this.getViewerController().mapComponent.getMap().addListener(viewer.viewercontroller.controller.Event.ON_GET_FEATURE_INFO,this.onFeatureInfoStart,this);
         //listen to a extent change
         this.getViewerController().mapComponent.getMap().addListener(viewer.viewercontroller.controller.Event.ON_CHANGE_EXTENT, this.onChangeExtent,this);
-        return this;        
+        this.config.viewerController.addListener(viewer.viewercontroller.controller.Event.ON_EDIT_SUCCESS, this.featureEdited, this);
```
mprins: odd to use `this.config.viewerController` here while bove we use `this.getViewerController()`

mprins (commented):
does this also work when a feature is deleted? (need to check)

mprins:
~~needs backport to 5.2.x branch~~ seems to be part of v5.2.0-rc2

mprins:
I've rebased to current master and force-pushed this

mprins:
I tested this and it works for attribute edits, but fails to update the extended feature panel  when a feature is deleted.

When a feature is deleted the panel should be erased and the marker should be removed from the map

mprins (changes requested):
it should also work when features are deleted, see other comments

mtoonen (approved):

mprins (approved):

PR #1131, opened 2018-07-03
fix saving the retrieveUploads and uploadDocument option value

mprins:


mtoonen (approved):
Thanks for picking this up!

PR #1136, opened 2018-07-05
Fixes erroneous drawn geometry when buffering geometry

mtoonen:


mprins (approved):

mprins:
@mtoonen does ths require a 5.2.x backport?

PR #1141, opened 2018-07-10
Check if layer exists when traversing the tree. 

mtoonen:
Also check if they exist in LayerSwitch and SelectionModule


mprins (approved):
always a good idea to check for existence

PR #1142, opened 2018-07-10
Check if layer exists when traversing the tree [v5.2.x]

mtoonen:
Check if layer exists when traversing the tree. Also check if they exist in LayerSwitch and SelectionModule


mprins (approved):
backport of #1141

PR #1143, opened 2018-07-10
Do a null-safe check, which account for null equality.

mtoonen:


mvdstruijk (approved):
Looks good. This fix won't throw an NullPointerException when the first parameter('(attributes.get(i)).getType()') is null. Note that this function: "Objects.equals()"  only works with java 7 or higher.

Issue #1147, opened 2018-07-12
Attributelist is refreshed after zooming in


geertplaisier:
Attributelist is refreshed after zooming in with no Extent filter configured

mprins:
see #1148

PR #1148, opened 2018-07-12
Check if layers are changed before resetting store

geertplaisier:
Fix for #1147 
Issue is that OpenLayers fires a ON_LAYER_VISIBILITY_CHANGED event (at some point) when zooming in. The LayerSelector listens to this event and resets the store, triggering a change event.

Fix is to first check whether the layers are actually different, if not, ignore the event


mtoonen (approved):

mtoonen:
Also backported to v5.2.x

PR #1151, opened 2018-07-13
Allow specified scripts to fail

mtoonen:

PR #1152, opened 2018-07-13
Bug/oracle drop if exists dont work

mtoonen:
ScriptRunner parses the statements by a delimiter ";". So the BEGIN construct doesn't work.
Expanded the scriptrunner capabilities in #1151, and allowed the drop table script to fail.


mprins (approved):

mprins:
cherry-picked both commits onto the v5.2.x branch

PR #1156, opened 2018-07-20
Let the measuretool use the snapping controller

mtoonen:
The idea was to separate the inline measureclass from the OpenlayerMapComponent. Then we could get the vectorlayer and add it to the snappingcontroller.
I've made a mock object (Line 89 - 96) to add to the OpenLayersSnappingController, which gets added to the control.
However, no snapping occurs.


mprins:
script does not compile: https://travis-ci.org/flamingo-geocms/flamingo/jobs/406210790#L1654-L1661

PR #1159, opened 2018-07-26
Allow JS to be executed when checkbox is set

geertplaisier:
Ext changed the way Javascript is allowed in HTML. We have to explicitly set the scope to null to prevent code to be added to container scope instead of global, which is required to define functions inside HTML block


mprins (approved):
OK, learn something every day... :bomb:

mprins:
cherry-pick'ed to v5.2.x

PR #1161, opened 2018-07-26
Load active layer when popup is opened first time

geertplaisier:
Also keep track of changed layers when popup is changed and load changed layer when popup opens


mprins (approved):

mprins:
cherry-pick'ed to v5.2.x

Issue #1113, opened 2018-06-27
collapsable panel can not be made visible


cvanlith:
### Version
5.1.5

### Browser info
chrome 67

### URL
flamingo4.b3p.nl mobiel

### Where and/or What
in viewer with closed collapsable panel

### What happened
collapsable panel can not made visible

### What did you expect?
opening panel should be possible

### Reproduce
  1.  configure left menu as floating
  2.  configure right top menu as collapsable, default collapsed
  3.  go to viewer
  4.  try to open right panel by clicking triangle

mprins:
closed with #1162

PR #1162, opened 2018-07-26
Prevent overlapping of toolbar over right panels

geertplaisier:
Fixes issue #1113 
Issue is that with a floating left menu the toolbar is moved to the right which then in overlaps the right panel


mprins (approved):

mprins:
cherry-pick'ed to v5.2.x

PR #1163, opened 2018-07-26
Properly set/restore color after loading file

geertplaisier:
When a file is saved the color in the file is correct. When loading the file, the feature that is drawn is properly colored, but the color property on the Feature object is not properly set after the feature is drawn. For new features this is automatically the default color which is incorrect for loaded features


mprins (approved):

mprins:
cherry-pick'ed to v5.2.x

PR #1164, opened 2018-07-26
better http status logging for failing wmts capabilities requests

mprins:
because things were sort of silent when not using a proper url


mprins:
cherry-pick'ed to v5.2.x

PR #1168, opened 2018-07-27
Bugs/mantis#11570 duplicate layers in legend

geertplaisier:
When loading a bookmark the event ON_LAYERS_INITIALIZED gets fired twice. The legend should be reset before re-creating to prevent duplicate legends.

Also changed this in 2 other components where duplication could also occur


mprins (approved):
LGTM

mprins:
cherry-picked both to v5.2.x

PR #1173, opened 2018-07-30
test for the existence of uploads before trying to render a table [v5.2.x]

mprins:
If there are no uploads an empty table would be rendered which does not comply with FOP schema and thus results in an invalid PDF being generated.

see also: [mantis 11527](http://mantis.b3p.nl/view.php?id=11527) and and #1174


mprins:
Yes I know; lots of whitespace fixes, someone used tabs and someone used spaces, please use spaces from now on. Fortunately GH provides an option to ignore whitespace in diff views

PR #1183, opened 2018-08-17
Feat/extended edit

geertplaisier:
- New edit component which listens to ON_FEATURE_INFO event from map to edit clicked features directly
- Edit and new Extended Edit component can now be placed next to the map (left/right margin)


mprins (commented):
not tried but LGTM. 
A small question: can you have both in a map? if you wanted to or inadvertently do so? 
User/Admin documentation seems to be failing as well

mtoonen (approved):

PR #1186, opened 2018-08-24
Improvements for Edit/ExtendedEdit/ExtendedFeatureInfo

geertplaisier:
- Allow click radius to be configured for Edit and ExtendedEdit
- Fixes an issue where the title would disappear on next/previous in ExtendedFeatureInfo
- Apply 'Allow edit' configuration in ExtendedEdit component
- Improved Edit component, after editing, keep selection of Layer


mprins (approved):
4px seems a safe default

PR #1247, opened 2018-10-22
Improvements on i18n

geertplaisier:
Adds translations in JSP
Further improves translations in JS
Adds English translations for JSP & JS


mprins (approved):

mprins:
```
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   ChooseApplicationActionBeanTest.deleteOriginalMotherWhenComponentsMustFollowNewlyPublishedApplication:185
[ERROR]   ChooseApplicationActionBeanTest.testDeleteApplication:232
[ERROR]   WFSTypeNamingTest.addWFSService:189 Saving WFS attribute source failed.
[INFO] 
[ERROR] Tests run: 47, Failures: 3, Errors: 0, Skipped: 0
```

mprins:
@cvanlith  I rebased this to master which ~~may solve some~~ dit not solve the build issues.

Running `mvn install` should run all the testcases and give you a stack trace locally

not sure about the `WFSTypeNamingTest.addWFSService` failure, the `ChooseApplicationActionBeanTest` failures are both NPE's in using the resource bundle; so I guess the testcase should be updated to als set the bundle.

I think there is a design issue because the bundle should be required and not null; now `getBundle()` can return `null` which apparently it does for the cases in `ChooseApplicationActionBeanTest`

viewer-admin/src/main/java/nl/b3p/viewer/admin/stripes/ApplicationActionBean.java
```diff
@@ -63,6 +66,20 @@ public ActionBeanContext getContext() {
         return context;
     }
 
+    /**
+     * @return the bundle
+     */
+    public ResourceBundle getBundle() {
+        return bundle;
```
mprins, line 76: it the locale/bundle is not set it should return a default and not `null` which it does now and causes some testcases to fail (because they don't set a bundle).
otherwise bundle should be a required parameter

mprins (changes requested):

viewer-admin/src/main/java/nl/b3p/viewer/admin/stripes/ApplicationActionBean.java
cvanlith: default is set now plus resources relocated to maven compliant location

mprins (approved):

mprins:
All builds are green, ready for merge?

@flamingo-geocms/committers We should look into tooling to keep translations up2date. There are several github apps, Crowdin offers a free plan for open source projects; https://github.com/marketplace/crowdin/

PR #1251, opened 2018-10-29
Impr/load large wm ses

mtoonen:


viewer-config-persistence/src/main/java/nl/b3p/viewer/config/services/Layer.java
```diff
@@ -78,6 +78,13 @@
         DETAIL_ALL_CHILDREN,
         DETAIL_WMS_STYLES
     }));
+    
+    private static Set<String> allowedSrsList = new HashSet<>(Arrays.asList(new String[] {
+        "EPSG:28992", // RD
+        "EPSG:4326", // WGS84
+        "EPSG:4937", //ETRS89
+        "EPSG:900913" //google
```
mprins: this change should be noted in the CRS project and picked up there to be enhanced. As it is now, just having EPSG:28992 is sufficient

mprins (changes requested):

mprins (approved):

PR #1252, opened 2018-10-29
use https URLs for the b3p domain

mprins:
because we have a new Nexus server


mprins (commented):
merging, all builds are green

PR #1254, opened 2018-10-30
Use existing jsessionid when passing it to the printactionbean

mtoonen:


viewer/src/main/webapp/WEB-INF/classes/log4j.properties
```diff
@@ -3,7 +3,7 @@ logFile=geo-viewer.log
 
 log4j.rootLogger=INFO,file
 
-log4j.logger.nl.b3p=INFO
+log4j.logger.nl.b3p=DEBUG
```
mprins: probably not what you want really

mprins (changes requested):

mprins (approved):

PR #1255, opened 2018-11-02
Bug/removed layer gives error in application

mtoonen:


viewer-admin/src/main/java/nl/b3p/viewer/admin/stripes/GeoServiceActionBean.java
```diff
@@ -61,6 +61,7 @@
 import java.io.StringWriter;
 import java.net.URL;
 import java.util.*;
+import nl.b3p.viewer.config.app.StartLayer;
```
mprins, line 64: unused import

mprins (approved):

PR #1259, opened 2018-11-07
Set color on featurestyle when a new CombineImakeWkt is created for a buffer otherwise color is always red

### LINKED ISSUE #1092: Buffer tool always produces red buffers
### Version
Which version of flamingo do you use:  5.1.5

### Browser info

n/a

### URL

http://geo02.lan.vrln.nl/viewer/app/geoviewer

### Where and/or What

When using the buffer component it always draws buffers in red even when selecting a different colour eg. green

### What did you expect?

Buffers should show up on the map in the selected colour

### Reproduce
How can we reproduce this? 

  1. buffer "geboorde putten" with 500m and green

###
requested urls for green and red:

`GET /viewer/action/Buffer?bbox=191855.68,415268.72,194409.28,416443.04&serviceId=36&layerName=Geboorde_putten&width=1520&height=699&buffer=500&maxFeatures=250&color=99CC00 HTTP/1.1`
  
`GET /viewer/action/Buffer?bbox=191855.68,415268.72,194409.28,416443.04&serviceId=36&layerName=Geboorde_putten&width=1520&height=699&buffer=500&maxFeatures=250&color=FF0000 HTTP/1.1`

so the frontend seems to request the coulour just fine and apparently it gets lost in the backend

mvdstruijk:
Set color on featurestyle when a new CombineImakeWkt is created otherwise color is always red

close #1092

### DIFF
--- a/viewer/src/main/java/nl/b3p/viewer/stripes/BufferActionBean.java
+++ b/viewer/src/main/java/nl/b3p/viewer/stripes/BufferActionBean.java
@@ -292,7 +292,9 @@ public class BufferActionBean implements ActionBean {
                 Geometry g = (Geometry) f.getDefaultGeometry();
                 if (g!=null){
                     g = g.buffer(buffer);                
-                    wkts.add(new CombineImageWkt(g.toText()));
+                    CombineImageWkt cwkt  = new CombineImageWkt(g.toText());
+                    cwkt.getStyle().setFillColor(color);
+                    wkts.add(cwkt);
                 }
             }
         } finally {


mprins (approved):

mtoonen:
Don't think this is entirely correct. FeatureStyle _should_ have the correct color. This error is probably a result that the buffercomponent isn't compatible with the (newly implemented) FeatureStyle class.
So I think the buffercomponent should set the color correctly. You can look at the https://github.com/flamingo-geocms/flamingo/tree/feat/extendDrawingCapabilities branch. There i did some adjustments for the drawing component to be completely compatible with the FeatureStyle class

mtoonen (changes requested):
Don't think this is entirely correct. FeatureStyle should have the correct color. This error is probably a result that the buffercomponent isn't compatible with the (newly implemented) FeatureStyle class.
So I think the buffercomponent should set the color correctly. You can look at the https://github.com/flamingo-geocms/flamingo/tree/feat/extendDrawingCapabilities branch. There i did some adjustments for the drawing component to be completely compatible with the FeatureStyle class

mtoonen (approved):

PR #1260, opened 2018-11-08
fixed problems with uploads in feature report

KJLammers:
featurereport didn't work properly with upload functionality

### DIFF
--- a/viewer/src/main/webapp/WEB-INF/xsl/print/FeatureReport.xsl
+++ b/viewer/src/main/webapp/WEB-INF/xsl/print/FeatureReport.xsl
@@ -91,7 +91,7 @@
                                     </xsl:for-each>
 
                                     <!-- uploads, if any -->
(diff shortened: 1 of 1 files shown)


viewer/src/main/webapp/WEB-INF/xsl/print/FeatureReport.xsl
```diff
@@ -91,7 +91,7 @@
                                     </xsl:for-each>
 
                                     <!-- uploads, if any -->
-                                    <xsl:if test="count(//extra/info[@classname='feature']/root/attr/__UPLOADS__) &gt; 0">
+                                    <xsl:if test="//extra/info[@classname='feature']/root/attr/__UPLOADS__/*">
```
mprins, line 94: this is the functional change, apparently the print xml format changed you now always have an `<__UPLOADS__>` element if uploads is checked on the layer.

mprins (commented):
A little elaboration on the other changes would be nice

mtoonen (approved):

PR #1264, opened 2018-11-15
fix getting base locale bundle, fixes #1262

### LINKED ISSUE #1262: Language always english
Setting the language in the viewer-admin to dutch, always results in an english.
https://github.com/flamingo-geocms/flamingo/blob/master/viewer/src/main/java/nl/b3p/viewer/stripes/I18nActionBean.java#L29
This always returns the english version, eventhough L28 is nl_NL locale.

Maybe because the locale on my machine is en-US, it defaults to the en_US resourcebundle?

matthijsln:


matthijsln:
Added extra changes to correctly get base bundle everywhere, also in viewer-admin

mtoonen (approved):

PR #1266, opened 2018-11-22
fix proxied scale dependent getLegendGraphic request

mprins:
- [x] adds test data and a test case for a proxied, scale dependent getLegendGraphic request
- [x] fix

~~_ps. fix is not tested IRL, but based on analysis_~~ tested, see below: https://github.com/flamingo-geocms/flamingo/pull/1266#issuecomment-441562085

see also [mantis 12110](https://mantis.b3p.nl/view.php?id=12110)


mprins:
Er moet ook een backport naar stable/5.2.x zegt @KJLammers

mprins:
tested with https://flamingo4.b3p.nl/geoserver/test_omgeving_fla5/wms? antenne-register layer, in viewer-admin I see two dots - which is expected as here the scale param is missing from the legend request:
image

in the viewer I see one dot: 
image

mtoonen (approved):

PR #1269, opened 2018-11-26
Fixed snapping with WFS sources

mtoonen:
moved figuring out sorting property to separate method. Fixed snapping with WFS sources. Fixes mantis 11537


viewer/src/main/java/nl/b3p/viewer/util/FeatureToJson.java
```diff
@@ -475,8 +456,33 @@ private void setSortBy(Query q, List<String> propertyNames, String sort, String
             } else {
                 sortAttribute = sort;
             }
-            this.setSortBy(q,sortAttribute,dir);
+        } else {
+            /* Use the first property as sort field, otherwise geotools while give a error when quering
+         * a JDBC featureType without a primary key.
```
mprins: I know this is a verbatim copy, but the condition if for both JDBCFeatureSource as well as WFSDataStore

mprins:
```suggestion
         * a featureType without a primary key.
```

mprins (changes requested):

mprins (approved):

PR #1273, opened 2018-11-29
Bump GeoTools from 20.0 to 20.1, also bump related dependencies

mprins:
Also update some maven plugins and the Tomcat test dependency.

see: https://geotoolsnews.blogspot.com/2018/11/geotools-201-released.html


mtoonen (approved):
:+1:

Issue #978, opened 2018-01-25
nl.b3p.viewer.config.services.JDBCFeatureSource logs connection password on DEBUG loglevel


matthijsln:
```
2018-01-25 10:59:11 DEBUG nl.b3p.viewer.config.services.JDBCFeatureSource:nl.b3p.viewer.config.services.JDBCFeatureSource.createDataStore(JDBCFeatureSource.java:207) - Opening datastore using parameters: {schema=public, database=x, fetch size=50, port=5432, passwd=****, Expose primary keys=true, dbtype=postgis, host=x, Primary key metadata table=gt_pk_metadata, user=x}
```

Replace password with * or something

mprins:
All Flamingo FeatureSources do this, when logging is set to debug, by design so the connection params can be debugged. Maybe we should set this to "trace" level.

Note that in error messages the password _is) replaced so in normal running conditions this is not an issue

PR #1278, opened 2018-12-04
do not log WFS password when debug logging is enabled, also log passw…

matthijsln:
…ord as '*' characters as long as password


mtoonen (approved):

mprins:
close #978

PR #1291, opened 2018-12-13
Fixes for lgtm

mtoonen:


mprins (approved):
LGTM;

fixed alerts:

- 8 for Reference equality test of boxed types
- 1 for Spurious Javadoc @param tags
- 1 for Boxed variable is never null
- 2 for Unused variable, import, function or class
- 1 for Useless assignment to local variable

PR #1306, opened 2018-12-21
Don't proxy alternativelegends

mtoonen:


geertplaisier (approved):
Looks fine to me. Adding isAlternative: false to every instance is not really nessecary since part.isAlternative will also be false(y) when isAlternative is not defined.

Issue #1032, opened 2018-05-18
Refactor featurereport


mtoonen:
With PR #/1030 I've copied some code from the FeatureReportActionBean to the PrintActionbean. Rewrite/refactor the FeatureReportActionBean to make use of this code and remove duplicate code. Probably the xsl's have to be adjusted as well to accommodate the slightly different format of the Print XML.

PR #1321, opened 2019-01-10
Feat/printstylesheets

cvanlith:
new stylesheets for master (and 5.2.x)
FeatureReport.xsl is updated, but uses old code (see #1032)


mprins:
too many conflicts, replaced with #1322

PR #1322, opened 2019-01-10
new xsl templates with improved checks to create scale bar

cvanlith:
and calculations to limit number of columns
new templates for A1 and A2

PR #1323, opened 2019-01-10
new xsl templates with improved checks to create scale bar

mprins:
and calculations to limit number of columns

backport of #1322 (and #1318 which fixes up travis)


mprins:
rebased to master so we get the updates of #1318 (which fixes the travis-ci builds)

mtoonen (approved):

PR #1334, opened 2019-01-18
use param.debug for loading the OpenLayers-debug libs

mvdstruijk:
Typo


mtoonen:
This is intentional: often you want to have the sourcefiles of the components, but not the OL sources. Loading the not minified version of OL slows down the application a lot.

PR #1356, opened 2019-02-15
Feat/mob fixes

mtoonen:


viewer/src/main/java/nl/b3p/viewer/stripes/EditFeatureActionBean.java
```diff
@@ -385,6 +391,10 @@ public Resolution delete() throws JSONException {
 
         return new StreamingResolution("application/json", new StringReader(json.toString(4)));
     }
+    
+    protected JSONObject getJsonFeature(String feature){
```
mprins, line 395: this can be public if it's not supposed to be private

viewer/src/main/java/nl/b3p/viewer/stripes/EditFeatureActionBean.java
```diff
@@ -77,11 +78,11 @@
...
     private ApplicationLayer appLayer;
 
-    private Layer layer;
+    protected Layer layer;
 
-    private SimpleFeatureStore store;
+    protected SimpleFeatureStore store;
 
-    private JSONObject jsonFeature;
+    protected JSONObject jsonFeature;
```
mprins, line 85: I would prefer using public getter/setter in an overriding class instead of making these `protected`

viewer/src/main/java/nl/b3p/viewer/stripes/EditFeatureActionBean.java
```diff
@@ -193,7 +198,7 @@ public Resolution edit() throws JSONException {
                 json.put("success", Boolean.TRUE);
             } while(false);
         } catch(Exception e) {
-            log.error(String.format("Exception editing feature", e));
+            log.error(String.format("Exception editing feature", e),e);
```
mprins: The string doesn't have any format specifiers to format, it should just be
```
            log.error("Exception editing feature", e);
```

viewer/src/main/java/nl/b3p/viewer/stripes/EditFeatureActionBean.java
```diff
@@ -294,12 +299,13 @@ public Resolution saveRelatedFeatures() throws JSONException {
                         json.put("success", Boolean.TRUE);
                     } catch (Exception ex) {
                         log.error(String.format("cannot save relatedFeature Exception: ",ex));
```
mprins: missing format specifiers, just use `log.error("cannot save relatedFeature Exception: ",ex);`

mprins (changes requested):

viewer/src/main/java/nl/b3p/viewer/stripes/EditFeatureActionBean.java
mtoonen: I've got a class that inherits the EditFeatureActionBean, so this construction makes it somewhat more transparent than using a getter/setter

viewer/src/main/java/nl/b3p/viewer/stripes/EditFeatureActionBean.java
mtoonen: Fixed in commit

viewer/src/main/java/nl/b3p/viewer/stripes/EditFeatureActionBean.java
mtoonen: Fixed in b352f9c66

mprins (approved):

PR #1362, opened 2019-02-25
Add MS SQL Server 2017 as an attribute source

### LINKED ISSUE #1337: Support for MS SQL server as an attribute source
### Version
Which version of flamingo do you use: 5.3.x

### What happened
 Cannot choose a MS SQL server database as an attribute source

### What did you expect?
 It should be possible to add a MS SQL server 2017 database as an attribute source and edit data including geometry

mprins:
### ToDo
- [x] Add gt-jdbc-sqlserver, make it possible to enter a sql server url as an attribute source

### Done
I've successfully made an application using a SQL server 2017 database with some BGT data and featureinfo and editing geometry works.

close #1337

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -384,6 +384,11 @@
                 <artifactId>gt-jdbc-postgis</artifactId>
                 <version>${geotools.version}</version>
             </dependency>
+            <dependency>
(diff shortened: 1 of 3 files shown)


mtoonen (approved):
Very cool!

PR #1368, opened 2019-03-06
Fixed #1316: overridden some OL2 functions. Problem seemed to be some…

### LINKED ISSUE #1316: can not select drawn object to change properties
### Version
5.3.10

### Browser info
latest chrome

### URL
https://flamingo5.b3p.nl/viewer/app/Solparc-BAG/v1

### Where and/or What
edit component

### What happened
can not select drawn object to change properties

### Reproduce

  1.  draw two or more objects
  2.  try to select previous object
  3.  is not possible!

mtoonen:
… sort of scope problem: retrieving the feature from the renderer didn't use the correct vectorlayer, but always the last one (which was the dummy vector layer)


mprins (approved):
lgtm

PR #1371, opened 2019-03-06
WIP Split LineString bugs

mtoonen:
- [x] linestring.contains(other) doesn't make sense for LineStrings
- [ ] Both splitstrategies don't create two features. Any ideas?

I think splitting lines have never worked. Possibly implement it based on: https://github.com/omar711/JUMP-Mapping-Platform/blob/master/src/main/java/com/vividsolutions/jump/workbench/ui/cursortool/SplitLineStringsOp.java


mprins:
Only splitting of objects with 2 dimensions was implemented if I recall correctly, there was no use case for splitting 1D objects.

splitting results in 2+n geometries, the chosen strategy is irrelevant for the actual split, it determines the way the results are stored (either 1 edited and 1 or more new features or alle new features and deleted splitted featuture)

PR #1429, opened 2019-05-06
Highlighting geometries and showing multiple styles in TOC

mtoonen:


viewer/src/main/java/nl/b3p/viewer/stripes/SimplifyFeatureActionBean.java
```diff
@@ -0,0 +1,275 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
mprins, line 16: should have proper license header or none

viewer/src/main/java/nl/b3p/viewer/stripes/SimplifyFeatureActionBean.java
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    public void setSolrconfig(SolrConf solrconfig) {
+        this.solrconfig = solrconfig;
+    }
+    // </editor-fold>
+
+    @Override
+    public AuditMessageObject getAuditMessageObject() {
+        throw new UnsupportedOperationException("Not supported yet."); //To change body of generated methods, choose Tools | Templates.
```
mprins: I think this should not throw an unchecked exception, return `null` instead or do something useful or don't implement Auditable (through featureinfo or attribute list an audit event would hav been generated already I think) and I don't think the result data -being a derived product- qualifies as original data

viewer/src/main/java/nl/b3p/viewer/stripes/SimplifyFeatureActionBean.java
```diff
@@ -0,0 +1,275 @@
...
+                json.put("success", Boolean.FALSE);
+                String error = null;
+                Layer layer;
+                EntityManager em = Stripersist.getEntityManager();
+
+                if (appLayer != null) {
+                    layer = appLayer.getService().getLayer(appLayer.getLayerName(), em);
+
+                    if (layer == null) {
+                        throw new Exception(getBundle().getString("viewer.editfeatureactionbean.3"));
```
mprins: maybe not a good idea to resue other bean's resource keys

viewer/src/main/java/nl/b3p/viewer/stripes/SimplifyFeatureActionBean.java
```diff
@@ -0,0 +1,275 @@
...
+                    }
+
+                    if (layer.getFeatureType() == null) {
+                        throw new Exception(getBundle().getString("viewer.editfeatureactionbean.4"));
```
mprins: maybe not a good idea to resue other bean's resource keys

viewer/src/main/java/nl/b3p/viewer/stripes/SimplifyFeatureActionBean.java
```diff
@@ -0,0 +1,275 @@
...
+
+    @Override
+    public AuditMessageObject getAuditMessageObject() {
+        throw new UnsupportedOperationException("Not supported yet."); //To change body of generated methods, choose Tools | Templates.
+    }
+
+    @DefaultHandler
+    public Resolution simplify() {
+        JSONObject result = new JSONObject();
+        Transaction transaction = new DefaultTransaction("edit");
```
mprins: back from holidays... but for reading a feature geometry do we need a transaction?

viewer/src/main/webapp/viewer-html/common/viewercontroller/controller/Event.js
```diff
@@ -125,6 +125,13 @@ Ext.define("viewer.viewercontroller.controller.Event",{
          * @param node the node that is expanded
          */
         TOC_EXPANDED                            : "TOC_EXPANDED",
+        /**
+         * @field
+         * Occures when a lstyle is selected for a certain layer
```
mprins:
```suggestion
         * Occurs when a style is selected for a certain layer
```

viewer/src/main/webapp/viewer-html/components/Highlight-config.js
```diff
@@ -0,0 +1,7 @@
+/* 
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+
```
mprins: should have proper license header or none

viewer-admin/src/main/resources/ViewerResources.properties
```diff
@@ -858,6 +858,7 @@ js.viewer.admin.applicationtreelayer.55 = Maak uw keuze
 js.viewer.admin.applicationtreelayer.56 = Hoogte
 js.viewer.admin.applicationtreelayer.57 = Alleen keuze uit lijst
 js.viewer.admin.applicationtreelayer.58 = Geen lege waarde toestaan
+js.viewer.admin.applicationtreelayer.59 = Stylen
```
mprins:
```suggestion
js.viewer.admin.applicationtreelayer.59 = Stijlen
```

mprins (changes requested):
Some small remarks inline.

Also: does it work when the layer is proxied ( I think it should because [STYLES](https://github.com/flamingo-geocms/flamingo/blob/master/viewer/src/main/java/nl/b3p/viewer/stripes/ProxyActionBean.java#L189) is allowed)

viewer/src/main/java/nl/b3p/viewer/stripes/SimplifyFeatureActionBean.java
mprins: Should not return `null` either, returning `null` will cause NPE.

mprins (approved):

PR #1433, opened 2019-05-08
don't set currentSearchID when there are more searchconfigs than 1

mvdstruijk:
https://mantis.b3p.nl/view.php?id=13594#c30467

/cc @KJLammers We need to know which version this is supposed to go? (previous stable - 5.2.3 or current stable 5.4.2)


mprins (commented):
I thought the cause of the issue was the access denied to the DBK sources and fixing this would not need a code change in Flamingo?...

mtoonen (approved):

PR #1437, opened 2019-05-10
Fixes #1412. Search for exact match

### LINKED ISSUE #1412: Default viewer cannot be set (sometime)
### Version
Which version of flamingo do you use: 5.4.0

### Browser info
latest chrome

### URL
https://flamingo5.b3p.nl/viewer-admin/action/chooseapplication/view

### Where and/or What
setting default viewer

### What happened
Trying to set Solparc-Bestemmingsplannen as defalt viewer always results in Solparc - BRK to be set as default. Some other viewers have the same problem, but not all.

### What did you expect?
setting of all viewers as default must be possible

geertplaisier:


mtoonen (approved):
lgtm

PR #1441, opened 2019-05-15
Bug/wmts printing

mtoonen:


mprins (approved):
makes sense to use format from service capabilities instead of relying on png being available

PR #1457, opened 2019-05-22
Use actionBeans['proxy'] global to set OpenLayers.ProxyHost

### LINKED ISSUE #1456: hardcoded actionbean paths used in various places, break proxy override
### Version
Which version of flamingo do you use:  master

### Where and/or What

I have an application with an override of the proxy url and it is largely ignored. This is the result of hardcoded proxy url in some places.

The FlamingoApploader construction may break the override mechanism that `app_overrides.jsp` provides for `actionBeans[]` array.
Also `SelectedContentCache.java` has a **hardcoded url path for the proxy**:

https://github.com/flamingo-geocms/flamingo/blob/634e73e19387711911c9b320b870a9e31573acb7/viewer-config-persistence/src/main/java/nl/b3p/viewer/util/SelectedContentCache.java#L140

### What happened

`getLegendGraphic` requests are routed through the main proxy, `getMap` requests are routed through the flamingo proxy.

it seems the getmap urls are modified in the SelectedContentCache to have the proxy prepended...

### What did you expect?

 `actionBeans["proxy"]=<js:quote><stripes:url beanclass="nl.b3p.viewer.stripes.OIGSProxyActionBean"/></js:quote>;`  should work for all components and actions

mprins:
- in identifytool
- SelectedContentCache: this works but is nasty :fire:

close #1456

### DIFF
--- a/viewer-config-persistence/src/main/java/nl/b3p/viewer/util/SelectedContentCache.java
+++ b/viewer-config-persistence/src/main/java/nl/b3p/viewer/util/SelectedContentCache.java
@@ -137,7 +137,9 @@ public class SelectedContentCache {
         int serverPort = request.getServerPort();
         String contextPath = request.getContextPath();
         StringBuilder url = new StringBuilder();
-        String servletPath = "/action/proxy/wms";
+        String servletPath = (request.getServletContext().getInitParameter("proxy") == null
+                ? "/action/proxy/wms"
+                : request.getServletContext().getInitParameter("proxy"));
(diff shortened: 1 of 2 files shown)


mprins:
another one: 
https://github.com/flamingo-geocms/flamingo/blob/634e73e19387711911c9b320b870a9e31573acb7/viewer-config-persistence/src/main/java/nl/b3p/viewer/util/SelectedContentCache.java#L140

mtoonen (approved):

PR #1458, opened 2019-05-24
Fixes saving required layers / auto-on layers for Solr Search config

geertplaisier:
mantis #13667

We should probably backport this fix to 5.4?


mtoonen (approved):

mtoonen:
Yes, can you  make a backport PR?

PR #1470, opened 2019-06-04
Create iterator object otherwise it will run forever

mvdstruijk:
Problem: The while loop will run forever if its true in the if-statement.

Solution: Make an Iterator object so the Iterator doesn't reset everytime it enters the while loop.


mtoonen (approved):

mprins (approved):
actually the problem is/was that there are two different iterators created once s part of the while and once inside the while,..

it's odd that the testcases don't catch this happening...

PR #1464, opened 2019-05-27
Added EditBulk.js component

tssmits:
Dear team,

This pull request adds a component called "Edit Bulk" that allows the user to select multiple elements, and change attribute values for all of them in one go. I have tested it on my local machine with a database that I was able to copy from a production environment.

I am curious to hear your remarks and am open to suggestions. Hopefully this pull request can be merged into master soon :)

PR #1486, opened 2019-06-29
Tests for EditFeatureActionBean and FeatureInfoActionBean

tssmits:
These tests have been moved here from https://github.com/flamingo-geocms/flamingo/pull/1464 because they are not directly relevant to that PR.

It depends on code from that PR, so right now, the CI tests will fail. As soon as that PR has been merged, I can rebase and the CI tests should pass.

### DIFF
--- a/viewer/src/main/java/nl/b3p/viewer/stripes/EditFeatureActionBean.java
+++ b/viewer/src/main/java/nl/b3p/viewer/stripes/EditFeatureActionBean.java
@@ -17,7 +17,6 @@
 package nl.b3p.viewer.stripes;
 
 import net.sourceforge.stripes.controller.LifecycleStage;
-import nl.b3p.viewer.util.AuditTrailLogger;
 import org.locationtech.jts.geom.Geometry;
 import org.locationtech.jts.io.WKTReader;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


mprins:
I've changed the access of `FeatureInfoActionBean#infoResponses()` and `EditFeatureActionBean#editResponse()` to package-private because these are not supposed to be called from outside the class and not designed to be overridden in a subclass.

Also removed the EditFeatureActionBean#addAuditTrailLog() as the actionbeans are not responsible for logging, only for creating the message which is what the `auditMessageObject` is for.

I'm not sure I like the (overridden) `getEntityManager()` method; in some other testcases we use dependency injection, so the entityManager is injected as an argument to the tested method eg. `editResponse()` would become `editResponse(EntityManager entityManager)`

Lets see what @mtoonen  thinks

tssmits:
Thanks for your feedback!

I haven't looked at this code for a while. I remember running into trouble with testing a stripes URL route "the proper way". Therefore I extracted the logic from edit() into editResponse() so that I could test that function instead. That caused problems with the EntityManager not being available, which I then chose to fix with the override method.

But it is all a bit hackish. There should be a proper way to unit test the edit() function directly, with the correct EntityManager automatically available.

I'll try and spend some time with the code this weekend, see if I can come up with a more elegant solution.

mtoonen:
Closed due to inactivity and irrelevance because of the featureapi

Issue #1499, opened 2019-07-11
multiple geotools EPSG authorities in (at least viewer) webapp 


mprins:
### Version
Which version of flamingo do you use:  master

### Where and/or What
 multiple geotools EPSG authorities in (at least viewer) webapp 

### What did you expect?
 There can be only one

PR #1500, opened 2019-07-11
Prevent having more than one GeoTools EPSG authority

mprins:
- [x] Exclude gt-epsg-hsql from commons-csw to prevent ending up with 2 instances of the EPSG factories
- [x] integration test
- [x] for current versions; exlude from wmts artifact
- [x] fix GeoTools (https://github.com/geotools/geotools/pull/2488 and https://github.com/geotools/geotools/pull/2487)

Note that when we next upgrade GeoTools either 21.x or 22.x we should revert 93146223ca1e4fbf875fab19eec23e56819eb7a8

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -288,6 +288,13 @@
                 <groupId>org.geotools</groupId>
                 <artifactId>gt-wmts</artifactId>
                 <version>${geotools.version}</version>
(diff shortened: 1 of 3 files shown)


mprins:
close #1499

PR #1523, opened 2019-08-16
add mutliple recipients/cc with comma separated string

mvdstruijk:
It is now possible to send emails to multiple recipients with a comma seperated string.
Example:
"email1@test.nl,email2@test.nl"


mprins (changes requested):
Kun je de javadoc ook aanpassen?

mprins (approved):

PR #1530, opened 2019-09-05
[BACKPORT 5.4.x]  Fixes using the selectionmodule after searching (and adding a layer t…

mtoonen:
…o the map with the searchresult). Restore the state of the added layer (it might be removed from startkaartbeeld) to visible

backport #1531


mprins (changes requested):
I think the changes to the context files should **not** be part of this bugfix; these are already defined in the web.xml.

mprins (approved):

PR #1536, opened 2019-09-09
Coordinate search functionality

mtoonen:
created config option for searching by coordinates. Implemented functionality in search component, also via url


mprins (approved):
LGTM

PR #1550, opened 2019-09-25
Enable "typeAhead" on combo in Edit component

### LINKED ISSUE #1549: "typeAhead" for Edit component combo's
### Is your feature request related to a problem? Please describe.
 
When a long list of limited possible values is available for a dropdown in the edit component it would be useful to "search" by typing a value, aka. typeAhead

### Describe the solution you'd like

enable typeAhead on combo's so they will search for matching values

### Describe alternatives you've considered

just scrolling through a list of more than 10 items gets painful

### Additional context

https://github.com/B3Partners/OIGS/issues/164

mprins:
Leave the `editable` option to default to `true`, set the `typeAhead` option to `true` and make the `forceSelection` listen to the configured option for being able to add arbitrary values or only values from the list.

close #1549 
close https://github.com/B3Partners/OIGS/issues/164

### DIFF
--- a/viewer/src/main/webapp/viewer-html/components/Edit.js
+++ b/viewer/src/main/webapp/viewer-html/components/Edit.js
@@ -1061,7 +1061,18 @@ Ext.define("viewer.components.Edit", {
             valueField: 'id',
             allowBlank: !disallowNull,
             disabled: !this.allowedEditable(attribute),
-            editable: !(attribute.hasOwnProperty('allowValueListOnly') && attribute.allowValueListOnly),
+            /* false to prevent the user from typing text directly into the field; the field can only have its value
+             set via selecting a value from the picker
+             (default: true) */
+            // editable: !(attribute.hasOwnProperty('allowValueListOnly') && attribute.allowValueListOnly),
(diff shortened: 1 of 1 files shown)


viewer/src/main/webapp/viewer-html/components/Edit.js
```diff
@@ -1062,6 +1062,8 @@ Ext.define("viewer.components.Edit", {
             allowBlank: !disallowNull,
             disabled: !this.allowedEditable(attribute),
             editable: !(attribute.hasOwnProperty('allowValueListOnly') && attribute.allowValueListOnly),
+            // typeAhead can only be enabled when editable is (also) true
+            typeAhead: true,
```
mprins, line 1071: We could also use the same value as `editable` `!(attribute.hasOwnProperty('allowValueListOnly') && attribute.allowValueListOnly),`

mprins:
after some digginf in the docs thare appears to be a property `forceSelection` to limit choices to the available list.

mtoonen (approved):
nice addition!

PR #1554, opened 2019-10-04
updated solr-core

mtoonen:


mprins (changes requested):
this doesn't compile / needs work

PR #1555, opened 2019-10-04
Use correct stringutils lib

mtoonen:


mprins (approved):

mprins:
~~you probably want to rebase after #1556 is merged so that testcases pass~~ rebase to master after #1556

PR #1572, opened 2019-10-17
Put the restricting query in the fq field for solr, so it will not be…

mtoonen:
… hindered by faulty/erroneous/vulnerable  userinput


mprins (approved):

mprins:
Maybe als create a [Security Advisory](https://github.com/flamingo-geocms/flamingo/security/advisories)?

PR #1577, opened 2019-10-18
Precisionmodel limited to 2 decimals. Unioned buffered geometries

mtoonen:


mprins (changes requested):
cdbf2ebc0289aebd84424d483af934efb0865a44 and 4a9deb65691f8daae551bfb516d98d9bfd499219 must be removed from this PR, perhaps you can do an interactive rebase and squash this PR into 2 distinct commits, one with the funcytional change and 1 with the version meddling/release

Not opposed to changing the precisionmodel, but would be nice to also fix the bug hidden away in there (#1256 oversight I guess), having hardcoded epsg 28992/RD, while we now also support WGS84

For decimal degrees I think 2 decimals is too large (2 decimal places ~~800m around here , see https://en.wikipedia.org/wiki/Decimal_degrees)

mprins:
this has gone stale after 15 months and many releases, also no bug or ticket describing the problem, so closing

PR #1623, opened 2019-12-10
Bug/featuretypes dont get deleted on feature source update

mtoonen:


viewer-config-persistence/src/main/java/nl/b3p/viewer/config/services/UpdatableFeatureSource.java
```diff
@@ -93,6 +93,7 @@ public FeatureSourceUpdateResult update(EntityManager em) throws Exception{
                 }
                 if(!stillExists){
                     it.remove();
+                    em.remove(oldFt);
```
mprins, line 96: is that all?... :smiling_imp:

viewer-config-persistence/src/main/resources/scripts/delete_orphan_featuretypes.sql
```diff
@@ -0,0 +1,9 @@
+
+delete from application_layer_attributes where attribute_ in (select id  from configured_attribute where feature_type in (select id from feature_type where id not in (select feature_type from feature_source_feature_types)));
```
mprins, line 2: do these updates+ deletes work on oracle + sqlserver; we have some oracle users and at least sqlserver user (for which  I'm pretty sure this problem does not exist, but I we can try on the OIGS database in-house

mprins (commented):
question if this upgrade script works on sqlserver (and/or oracle) me or @mvdstruijk  can try thursday on inhouse OIGS database - which I'm pretty sure does not have these issues.

mprins:
ps. if you want the auto-backport to work you need to use the squash and merge button and not create a merge commit.
I changed the label since this PR targets master

viewer-config-persistence/src/main/java/nl/b3p/viewer/config/services/UpdatableFeatureSource.java
mtoonen: I don _not_ want to talk about it.

viewer-config-persistence/src/main/resources/scripts/delete_orphan_featuretypes.sql
mtoonen: Oracle works

mprins (approved):
mssql works as well

PR #1647, opened 2020-01-10
Raise base level of z-indices, so other layers don't interrupt with v…

mtoonen:
…ectorlayers.

Previously found workaround for agel


matthijsln (approved):

mprins:
@mtoonen  Does this need to be backported to 5.6.x / 5.4.x?

PR #1655, opened 2020-01-21
fix session id exposure in DownloadMap

### LINKED ISSUE #1654: DownloadMap component exposes session id
### Version
Which version of flamingo do you use: 5.7.2-SNAPSHOT

### Where and/or What

When using the "Download Map:" component (which uses the CombineImageActionBean with GET requests) the session id is exposed in a browser url

### What happened
Session id's are exposed in browser url creating a possible sensitve data leak.

Note that the print previes also uses a GET to get the preview image which exposes the url in the "network"  log facility of any browser developer tools

AFAICT The "other" browser is not logged in

### What did you expect?
 Should not be possible to re-use the url

### Reproduce
How can we reproduce this? 

  1. create an application with login+secured service and "download map" component
  2. login to application and download the map
  3. use the shown download link in an anonymous browser window (not logging in)

mprins:
- fix line-endings for CombineImageActionBean
- Use POST for getting preview image
- Use a POST request to a download endpoint to download the map

Also deprecate unsafe methods

close #1654


matthijsln (changes requested):
Download map works, but not in IE11:
```
SCRIPT5: Access is denied.
DownloadMap.js (77,9)
```
at line with ` link.click();`

 The `getImage()` handler is still used by the print preview, if that is changed the `getImage()` method and the duplicate code shared with `download()` can be removed. The print preview URL still contains the session ID. It's not that easy to leak it as with sharing the download URL, but it is still exposed to proxies, so if it can be converted to a POST that would be better.

The `Expires` header can be removed, caching is not working anyway (don't think it did with GET before).

matthijsln:
Edge 18 and 79 work, so depends on if IE11 is still supported. Workaround small but pretty ugly, see https://stackoverflow.com/a/37522444

mprins:
not the right time to drop IE11 yet... will fix

mtoonen (approved):
lgtm

PR #1664, opened 2020-01-28
update error codes to trap in the generic page

mprins:
checked `ErrorResolution`s and I see the following codes

- [x] SC_FORBIDDEN / 403
- [x] SC_NOT_FOUND / 404
- [x] SC_INTERNAL_SERVER_ERROR / 500

Remote services may also pass 
- [x] SC_SERVICE_UNAVAILABLE / 503
- [x] SC_BAD_REQUEST / 400

and session expiry can cause 
- [x] SC_UNAUTHORIZED / 401

followup of #1653


matthijsln (approved):
In Servlet 2.5 there is no catch-all so unfortunately have to list every possible code, hard when passing through other servers' responses... 

See this discussion: https://stackoverflow.com/a/7066536

I guess a servlet filter can be used as a catch-all, but that would be a bit excessive to do and just clutter code for no real benefit.

PR #1683, opened 2020-02-14
Support for geofenceheader

mvdstruijk:
It is now possible to set a geofenceheader key on a geoservice. This is only applied when proxy is used. 
Instructions:
- Set geofenceheader key (this is the key that has been set in geoserver with geofence) on the geoservice and set use proxy.
- Make a group/role in the viewer-admin with description value 'geofence', the group/role name must match a geofence rule.

### DIFF
--- a/viewer-config-persistence/src/main/java/nl/b3p/viewer/config/services/GeoService.java
+++ b/viewer-config-persistence/src/main/java/nl/b3p/viewer/config/services/GeoService.java
@@ -59,7 +59,7 @@ public abstract class GeoService implements Serializable {
      * authentication.
      */
     public static final String PARAM_PASSWORD = "password";
-
+    
     @Id
     private Long id;
 
@@ -74,7 +74,9 @@ public abstract class GeoService implements Serializable {
 
     private String username;
     private String password;
-
+    
+    private String geofenceHeader;
+    
     private boolean monitoringEnabled;
     
     private boolean monitoringStatusOK = true;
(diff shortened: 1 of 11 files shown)


viewer-config-persistence/src/main/java/nl/b3p/viewer/config/services/GeoService.java
```diff
@@ -74,7 +74,9 @@
 
     private String username;
     private String password;
-
+    
+    private String geofenceHeader;
```
mprins, line 78: I think this adds an extra columnt in the datamodel, no?
In that case upgrade scripts must be provided to do that, see https://github.com/flamingo-geocms/flamingo/tree/master/viewer-config-persistence/src/main/resources/scripts

mprins (changes requested):
I think we need some SQL as well

viewer/src/main/java/nl/b3p/viewer/stripes/ProxyActionBean.java
```diff
@@ -149,7 +151,22 @@ private Resolution proxyWMS() throws IOException, URISyntaxException {
 
         HttpClientConfigured client = getHttpClient(theUrl, em);
         HttpUriRequest req = getHttpRequest(theUrl);
-
+        
+        GeoService gs = em.find(GeoService.class, serviceId);
+        if(gs.getGeofenceHeader() != null ){
+            User user = em.find(User.class, context.getRequest().getUserPrincipal().getName());
```
mprins, line 157: this might not work when printing

viewer-config-persistence/src/main/java/nl/b3p/viewer/config/services/GeoService.java
mvdstruijk: Yes you are right.

viewer/src/main/java/nl/b3p/viewer/stripes/ProxyActionBean.java
mvdstruijk: When i tested this it worked.

mprins (commented):
De SQL Server variant van het script ontbreekt nog

mprins (approved):

PR #1689, opened 2020-02-21
Include the Oracle ojdbc8 database driver in the Tomcat lib package

mprins:
As this is available from Maven Central under the '_Oracle Free Use Terms and Conditions_'

see: https://search.maven.org/artifact/com.oracle.ojdbc/ojdbc8/19.3.0.0/jar


mtoonen (approved):
Nice!

PR #1696, opened 2020-02-24
Limit the number of features to be downloaded

mtoonen:


mprins (approved):
:heavy_check_mark:  OK, but would prefer to have this as a context param, so it can be tuned in deployment

PR #1701, opened 2020-02-24
IE11 hack for undefined error when you have simple filters that are set during startup

mprins:
see [mantis-14365](https://mantis.b3p.nl/view.php?id=14365)


mtoonen (approved):
Time for IE11 to die...

PR #1720, opened 2020-03-04
Only use last part of FID (the actual value of the PK). featuretypena…

mtoonen:
…me.value does not work for materialized views

I'm a bit confused as to why this works... Maybe @mprins  knows?

### DIFF
--- a/viewer/src/main/java/nl/b3p/viewer/stripes/FeatureReportActionBean.java
+++ b/viewer/src/main/java/nl/b3p/viewer/stripes/FeatureReportActionBean.java
@@ -139,7 +139,7 @@ public class FeatureReportActionBean extends LocalizableApplicationActionBean im
             FilterFactory2 ff = CommonFactoryFinder.getFilterFactory2();
             FeatureSource fs = layer.getFeatureType().openGeoToolsFeatureSource(TIMEOUT);
             String geomAttribute = fs.getSchema().getGeometryDescriptor().getLocalName();
-            Filter fidFilter = ff.id(new FeatureIdImpl(this.fid));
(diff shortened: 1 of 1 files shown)


viewer/src/main/java/nl/b3p/viewer/stripes/FeatureReportActionBean.java
```diff
@@ -139,7 +139,7 @@ public Resolution print() throws URISyntaxException, IOException, Exception {
             FilterFactory2 ff = CommonFactoryFinder.getFilterFactory2();
             FeatureSource fs = layer.getFeatureType().openGeoToolsFeatureSource(TIMEOUT);
             String geomAttribute = fs.getSchema().getGeometryDescriptor().getLocalName();
-            Filter fidFilter = ff.id(new FeatureIdImpl(this.fid));
+            Filter fidFilter = ff.id(new FeatureIdImpl(this.fid.substring(this.fid.indexOf(".")+1)));
```
mprins: so what happens when typename contains a dot?

use `this.fid.replace(layer.getFeatureType().getTypeName() + ".", "")`

has the view been properly registered in the metadata table?

mtoonen: "so what happens when typename contains a dot?"
We should name-and-shame that person... 
I'll edit it so it'll use lastIndexOf...

View is present in gt_pk_metadata table...

mprins: I still think it's better to use `this.fid.replace(layer.getFeatureType().getTypeName() + ".", "")` instead of just getting the part of the string after the last dot

mprins (approved):
I still think this is odd... so 50/50 on merging

PR #1723, opened 2020-03-05
Feat/added expire date for user

mvdstruijk:
It is possible to set an expiry-date on a user.

This date is used for login to a viewer.

If there is no date set on the user, the user is allowed to login.


viewer/src/main/java/nl/b3p/viewer/stripes/ApplicationActionBean.java
```diff
@@ -411,7 +414,30 @@ public static Resolution checkRestriction(ActionBeanContext context, Application
...
+        }
+
+        if(today.after(expire)){
+            ResourceBundle bundle = ResourceBundleProvider.getResourceBundle(determineLocaleForBundle(context, application));
+            String msg = bundle.getString("viewer.applicationactionbean.expired");
+            context.getValidationErrors().addGlobalError(new SimpleError(msg));
+            context.getRequest().getSession().invalidate();
+            return new ForwardResolution("/WEB-INF/jsp/error.jsp");
+        }
+        else if (!Authorizations.isApplicationReadAuthorized(application, context.getRequest(), em) && (username == null || u != null && u.isAuthenticatedByIp())) {
```
mprins: I think it would be better to place this in [Authorizations](https://github.com/flamingo-geocms/flamingo/blob/master/viewer-config-persistence/src/main/java/nl/b3p/viewer/config/security/Authorizations.java)

possibly as part of `isReadAuthorized`, maybe a a new method

the parseError should not go to the user.
the login failure should be a generic login failure error and redirect to the loginfail jsp/page, not the error page

mprins (commented):
let's see what @mtoonen thinks

mprins:
@mvdstruijk if this needs to be backported to 5.6.x:
- add label "backport v5.6.x"
- use "squash merge" option when merging

and github will attempt to create a new PR for the v5.6.x branch

viewer/src/main/java/nl/b3p/viewer/stripes/ApplicationActionBean.java
mvdstruijk: Yes its better to move the code to its own method in Authorizations.
Not sure if redirecting to the loginfail page is better, because the user gets a default message that the login information is incorrect what is not the case here. Besides the user can try to login again, this is useless because the account has expired.

mtoonen (approved):

viewer-config-persistence/src/main/java/nl/b3p/viewer/config/security/Authorizations.java
```diff
@@ -106,6 +117,8 @@
     private static final Object LOCK = new Object();
     
     private static final String ROLES_ATTRIBUTE = Authorizations.class.getName() + ".roles";
+
+    private static final Log log = LogFactory.getLog(ScriptRunner.class);
```
mprins:
```suggestion
    private static final Log log = LogFactory.getLog(Authorizations.class);
```

mprins (approved):

PR #1729, opened 2020-03-10
Forwardport/stable bugfixes

mtoonen:
bugfixes from 5.6 stable release


mprins (approved):
OK with me; kind of impossible to get what the problem being fixed for a commit is without problem ticket

mprins:
@mtoonen 
There's a problem in this commit db15d4fe6280b805b798ed31e50c51b38bcad5d3

```
07-Apr-2020 13:28:02.676 SEVERE [http-nio-7081-exec-3] com.google.javascript.jscomp.LoggerErrorManager.println /mnt/tera1/home/mark/dev/projects/flamingo-ibis/viewer/target/ROOT/viewer-html/components/Snapping.js:264: ERROR - this language feature is only supported for ECMASCRIPT6 mode or better: arrow function. Use --language_in=ECMASCRIPT6 or ECMASCRIPT6_STRICT or higher to enable ES6 features.
        this.layerList.forEach(l =>{
```

You cannot use arrow functions in IE11... 

I think

https://github.com/flamingo-geocms/flamingo/blob/b0ddbaf4aa386ef601dd78fcb82b4754858df8e2/viewer/src/main/webapp/viewer-html/components/Snapping.js#L264-L266

should be something like

```js
        this.layerList.forEach(function(l) {
           layers.push(l.id);
        });
```

currently we are at ECMASCRIPT3 (if any specified):  https://github.com/flamingo-geocms/flamingo/blob/b0ddbaf4aa386ef601dd78fcb82b4754858df8e2/viewer/pom.xml#L350

PR #1754, opened 2020-03-19
isAlternative = true from appLayer legends

mtoonen:
[mantis-13539](https://mantis.b3p.nl/view.php?id=13539)

Legend images configured in boomstructuur met kaarten, weren't identified as alternative legends, and were rerouted through the proxy. Which declined the request.


mprins (approved):
LGTM, 

out of curiosity; does this work when printing?

mtoonen:
It even worked in printing without this fix

PR #1772, opened 2020-03-26
WIP Feat/upgrade hibernate

mtoonen:
- [x] Upgrade libs
- [x] schemaexport
- [x] Unit tests passen
- [ ] integratie tests
- [ ] schema migratie SQL Server
- [ ] schema migratie Oracle (zie bijv. https://docs.oracle.com/en/database/other-databases/nosql-database/19.5/java-driver-table/altering-or-dropping-identity-column.html)


mprins:
I would prefer to keep things the same as much as possible so upgrading is as painless as possible... 

## schema incompatiblilities for TextType

eg. old / oracle:
`configured_component.config clob`
new: 
`configured_component.config long`

(note that LONG should be avoided and you can only have 1 LONG per table)

for sql server some things are also going askew..
old:
`configured_component.config text null`

new:
`configured_component.config varchar(MAX)`

Whereas the postgresql verion is `configured_component.config text` for both old and new.

FWIW the java code
```java
    @Lob
    @Type(type = "org.hibernate.type.TextType")
    private String config;
```

it may seem that the Oracle dialect mapping `java.lang.String` to `LONG` 
https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#_jdbc_related_options

from: https://docs.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql?view=sql-server-ver15
> IMPORTANT! ntext, text, and image data types will be removed in a future version of SQL Server. Avoid using these data types in new development work, and plan to modify applications that currently use them. Use nvarchar(max), varchar(max), and varbinary(max) instead.

So I guess we can't upgrade automatically when we implement these changes

mprins:
@mtoonen Ik denk dat deze is ingehaald door #2670 en dus gesloten kan worden?

PR #1790, opened 2020-04-07
Allow passing extra parameters to remove() and edit() functions

mprins:
Allow passing extra parameters to `remove()` and `edit()` functions of EditFeature, similar to FeatureInfo requests.

needed for https://github.com/B3Partners/flamingo-ibis/issues/137


mtoonen:
Looks fine, but can't we already define what the parameters are going to be? It would make the contract a bit more clear?

mtoonen (approved):

mprins:
> Looks fine, but can't we already define what the parameters are going to be?

in this case some flags for a custom edit actionbean

PR #1791, opened 2020-04-07
remove arrow function from snapping component

mprins:
also up the ecmascript version of the compiler config

see also https://github.com/flamingo-geocms/flamingo/pull/1729#issuecomment-610354951


mtoonen (approved):
LGTM. Ik ben af en toe mijn tijd ver vooruit :P

PR #1800, opened 2020-04-17
If dynamic lookup is configured for editing an attribute do not put out (possibly old) editValues

### LINKED ISSUE #1799: Edit component does not pick up dynamic attribute list when a static lsit was defined before
### Version
Which version of flamingo do you use:  5.7.3-SNAPSHOT

### What happened

When a static lookup list is defined for editing an attribute and after that it is switched to a database lookup the edit component does not pick up the dynamic lookup becasue the `editValues` is still pouplated with the old (static) values.

### What did you expect?
Edit component should use configured mechanism (static or dynamic) and not the `editValues` list

### Reproduce

see above

mprins:
close #1799


mtoonen (approved):
LGTM

PR #1809, opened 2020-05-06
geometry_columns do not exist in oracle

mtoonen:
Is this correct, or does sqlserver also doesn't have a geometry_columns table/view?


mprins:
"geometry_columns" is a GeoTools (and GDAL) thing, it should be created and populated in the datastore, just like a "gt_pk_metadata" that can be used to have fixed feature-id values when publishing views.
 
and not just for oracle, but also for postgis and sql server, see eg. https://docs.geoserver.org/latest/en/user/data/database/postgis.html#publishing-a-postgis-view

since there is no bug description, not sure what this should fix, but is oracle breaks on this and others don't (even when that table is missing) this should be fixed in geotools

mtoonen:
It raises an exception when an attributesource doesn't have the geometry_columns table. You said it was necessary to add for SQLServer, and postgis has it by defintion, so this check should suffice I think

mprins:
it's needed to determine the geometry type if there are no records in a table (because neither sqlserver nor oracle have specific types) 
I don't remember if sqlserver works without and apparently this kind of thing is not part of the manual/user test suite :unamused: 

so I guess this is an OK workaround, perhaps we could add this as a param on the JBDC connection at some point in time, so it becomes user config

mprins (approved):
OK as a workaround, I think geotools shouldn't throw up over this, but it apparently does

PR #1815, opened 2020-05-19
Feature: improve attribution

matthijsln:
- Updated suggested OSM attribution with direct link to copyright page

- Improved legibility of OL2 attribution, before and after:

image
image

- Expanded OL5 attribution by default:

image

Previously:
image


mtoonen (changes requested):
Does it also work in OL5?

matthijsln:
> Does it also work in OL5?

Yes, I also changed the default attribution state from collapsed to uncollapsed when using OpenLayers5Map, as you can see in the screenshots.

mtoonen (approved):

PR #1816, opened 2020-05-19
backport from @mprins. Arrowfunctions not in ie11, so revert this epi…

mtoonen:
…c functionality to support pre-historic browsers


mprins (approved):
LOL! DeepCode says: 
> Use an arrow function inside forEach() instead of an anonymous function. For anonymous functions the value of 'this' may differ depending on the calling context, which often leads to unexpected errors. On contrary, the value of 'this' used in an arrow function is determined by the scope where the function is defined.
https://www.deepcode.ai/app/gh/flamingo-geocms/flamingo/20d5ff83d8b37557a21c9e6792b0b10811a60b3b/flamingo-geocms/flamingo/3c11ab47ed4f2ad1566628f783a21d4f0da3cff9/pr/_/%2F/code

PR #1858, opened 2020-06-22
Allow setting the label width of attribute labels on the edit form

mprins:
the default value is 100, which is often too small, this will allow setting attribute label width for each instance of the edit component.

alternatively one could add css `.editCmpLbl {width: 250px !important;}` but that would apply applicationwide..


mprins:
I don't need a backport

mtoonen (approved):
Seems ok

PR #1859, opened 2020-06-22
make the window that holds the list of edit features wider

mprins:
make the window that holds the list of edit features to choose from 80% wide instead of 500px so it is more useful on a big and/or hi-res screen

(I don't need a backport)


mtoonen (changes requested):
80% seems a bit much for a 4K monitor. Maybe also add a maxWidth?

mtoonen (approved):

Issue #1665, opened 2020-01-28
CombineImageActionBean leaks sessionid


mprins:
### Version
Which version of flamingo do you use: 5.7.2-SNAPSHOT

### Where and/or What
viewer, eg. print preview

### What happened
 sessionid is part of the preview image url, the url could be snooped and reused, perhaps for a session hijack

see also: 

- #1654
- #1655
- https://github.com/flamingo-geocms/flamingo/pull/1655#pullrequestreview-349242987

PR #1882, opened 2020-06-29
use cookie for session tracking

matthijsln:
This does restrict the very few people who disable first-party cookies from using apps requiring authentication, but public apps should still be useable.

### DIFF
--- a/viewer/src/main/webapp/WEB-INF/web.xml
+++ b/viewer/src/main/webapp/WEB-INF/web.xml
@@ -210,6 +210,7 @@
     </login-config>
     <session-config>
         <session-timeout>240</session-timeout>
+        <tracking-mode>COOKIE</tracking-mode>
     </session-config>
     <welcome-file-list>
         <welcome-file>index.jsp</welcome-file>


matthijsln:
May interfere with https://github.com/flamingo-geocms/flamingo/blob/master/viewer/src/main/java/nl/b3p/viewer/print/PrintUtil.java#L90
Although that code should be changed anyway, see #1655

mprins:
may also interfere with printing secured layers through proxy

mtoonen:
> May interfere with https://github.com/flamingo-geocms/flamingo/blob/master/viewer/src/main/java/nl/b3p/viewer/print/PrintUtil.java#L90
> Although that code should be changed anyway, see #1655


> may also interfere with printing secured layers through proxy

Both seem to work

mtoonen (approved):

mprins:
this needs extra care for war-overlay deployments, as these may have their own `web.xml`

Also this doesn't resolve the session id exposure described in #1665, I can still see a request for http://kantoor.b3p.nl:88/action/combineimage?getImage=t&imageId=KCFW5G007O&JSESSIONID=2F95ACBF4BC22B271485CBDF97C276EB that exposes the session id when loading the preview in the print component.

mtoonen:
> this needs extra care for war-overlay deployments, as these may have their own `web.xml`

People shouldn't do that :-p

> 
> Also this doesn't resolve the session id exposure described in #1665, I can still see a request for http://kantoor.b3p.nl:88/action/combineimage?getImage=t&imageId=KCFW5G007O&JSESSIONID=2F95ACBF4BC22B271485CBDF97C276EB that exposes the session id when loading the preview in the print component.

Not sure if this is at all fixable...

PR #1886, opened 2020-07-02
Fix Oracle syntax error in upgrade script

mprins:


mtoonen (approved):
Oooo, thats why upgrades failed... Thanks!

mprins:
thanks @BaderAhmad

mprins:
I think maybe we should set up automated testing for schema upgrades. Did something similar for https://github.com/B3Partners/brmo

mprins:
does not exist in v5.4.x

PR #1898, opened 2020-07-09
Bug/performance layer list

mtoonen:
Fix for slow ajaxrequests in Gouda. This moves that functionality to the front-end.

### DIFF
--- a/viewer-config-persistence/src/main/java/nl/b3p/viewer/util/databaseupdate/DatabaseSynchronizer.java
+++ b/viewer-config-persistence/src/main/java/nl/b3p/viewer/util/databaseupdate/DatabaseSynchronizer.java
@@ -174,6 +174,11 @@ public class DatabaseSynchronizer implements Servlet {
         
         // 5.7.3
         updates.put("39", new UpdateElement(Collections.singletonList("add_geoservice_geofence_header.sql"), String.class, true));
+
+        //5.6.4
+
+        updates.put("40", new UpdateElement (Collections.singletonList("selectedcontentcaches_dirty.sql"), String.class));
+
         // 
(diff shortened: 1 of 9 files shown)


viewer-config-persistence/src/main/java/nl/b3p/viewer/util/databaseupdate/DatabaseSynchronizer.java
```diff
@@ -174,6 +174,11 @@
         
         // 5.7.3
         updates.put("39", new UpdateElement(Collections.singletonList("add_geoservice_geofence_header.sql"), String.class, true));
+
+        //5.6.4
+
+        updates.put("40", new UpdateElement (Collections.singletonList("selectedcontentcaches_dirty.sql"), String.class));
```
mprins, line 180: this file is missing from this PR

mprins (changes requested):
seems therr is a file missing. 
Also what are expected problems on downstream projects? eg oigs, ibis, ... that have custom components. a quick note would be good on API changes

viewer-config-persistence/src/main/java/nl/b3p/viewer/util/databaseupdate/DatabaseSynchronizer.java
mtoonen: It's an older, already existing file which I reused

mtoonen:
> seems therr is a file missing.
> Also what are expected problems on downstream projects? eg oigs, ibis, ... that have custom components. a quick note would be good on API changes

Not really an API change: backend didn't change (thought about removing the LayerListActionBean, but decided against it for these reason). Everything stayed the same, except for no ajaxcalls for retrieving the layerlist

mprins (approved):

PR #1899, opened 2020-07-09
configuration option for the maximum amount of features in the download functionality of the attributelist

mtoonen:
default is 1000


mprins (approved):
:tada:

PR #1967, opened 2020-08-31
Feat/gbiw 80 fix add feature button

mtoonen:


mprins (approved):
lgtm, I'm assuming `startGeometryDrawing` is not an Extjs event but an Angular event

PR #1971, opened 2020-09-02
Dev/gbiw 141 tslint

matthijsln:
I did not run `npm run lint-fix` yet, because of GBIW-140.

### DIFF
--- a/viewer/src/main/webapp/flamingo-components/package.json
+++ b/viewer/src/main/webapp/flamingo-components/package.json
@@ -8,6 +8,7 @@
     "build-prod": "npm run ng -- build --prod --output-hashing=none",
     "test": "ng test --project=core",
     "lint": "ng lint --project=bridge",
+    "lint-fix": "ng lint --project=bridge --fix",
     "e2e": "ng e2e"
   },
   "private": true,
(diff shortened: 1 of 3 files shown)


viewer/src/main/webapp/flamingo-components/package.json
```diff
@@ -7,7 +7,8 @@
     "build": "ng build",
     "build-prod": "npm run ng -- build --prod --output-hashing=none",
     "test": "ng test --project=core",
-    "lint": "ng lint",
+    "lint": "ng lint --project=bridge",
+    "lint-fix": "ng lint --project=bridge --fix",
```
mtoonen, line 11: This should be project=bridge, but also core
bridge is the glue between angular and extjs, core contains the actual components

mtoonen:
Please use feat/GBIW-123_ for branchnames. This will make youtrack pick up commits

viewer/src/main/webapp/flamingo-components/package.json
matthijsln: bridge project also lints core files

matthijsln:
Superseded by #1973 1973

PR #1975, opened 2020-09-03
Feat/gbiw 136 bridge functionality ext and ng

mtoonen:
Bridge for interfacing between angular and extjs apps. Ideas/comments/suggestions?


matthijsln:
The principle I gather is:
- from ExtJS to Angular: set an attribute on the `tailormap-mapbridge` custom element
- from Angular to ExtJS: emit an event from the custom element, the ExtJS component listens to it and forwards it.

This can work, but the bridge may get very large when adding all possible events and API, especially because an attribute is only a string so objects must be stringified. Also when using different classes that would need a lot of mapping.

Because the web component (although packaged with webpack) still runs in the same JavaScript scope as the other JS / ExtJS code, we could also use objects directly. Flamingo defines the global `viewerController` variable. We could add a wrapper around that, maybe use (parts of) it directly and just provide typings.

mprins:
:confounded:
We don't have too many [events](https://github.com/flamingo-geocms/flamingo/blob/master/viewer/src/main/webapp/viewer-html/common/viewercontroller/controller/Event.js) and probably don't need all of them to carry over, but wrapping `viewer.viewercontroller.ViewerController` sounds like a good idea

mtoonen:
> The principle I gather is:
> 
> * from ExtJS to Angular: set an attribute on the `tailormap-mapbridge` custom element
> * from Angular to ExtJS: emit an event from the custom element, the ExtJS component listens to it and forwards it.
> 
> This can work, but the bridge may get very large when adding all possible events and API, especially because an attribute is only a string so objects must be stringified. Also when using different classes that would need a lot of mapping.
> 
> Because the web component (although packaged with webpack) still runs in the same JavaScript scope as the other JS / ExtJS code, we could also use objects directly. Flamingo defines the global `viewerController` variable. We could add a wrapper around that, maybe use (parts of) it directly and just provide typings.

Hmm, this sounds better. I'll work on that. Thanks!

matthijsln:
See https://github.com/flamingo-geocms/flamingo/pull/1984

PR #1985, opened 2020-09-07
Feat/gbiw 141 apply lint fix

matthijsln:


tailormap-components/package.json
```diff
@@ -7,8 +7,8 @@
     "build": "ng build",
     "build-prod": "npm run ng -- build --prod --output-hashing=none",
     "test": "ng test --project=core",
-    "lint": "ng lint --project=bridge",
-    "lint-fix": "ng lint --project=bridge --fix",
+    "lint": "ng lint --project=core",
+    "lint-fix": "ng lint --project=core --fix",
```
mtoonen: I thought that specifying core didn't lint the bridge project

mtoonen (approved):
Bit large to oversee all changes, but looks good and works

tailormap-components/package.json
matthijsln: True, but specifying only bridge did not lint all files from core. But if you don't specify a project, you get duplicate errors. Maybe that's better.

PR #2051, opened 2020-10-12
Get GBI releases working

mprins:
Rework viewer pom file; add a "clean" config so that the tailomap-components dist is cleaned (because it is also build using maven, it should also be cleaned using maven) This was actually hiding the cause of the problem.

The copy-resources config that copies the `tailormap/dist` into the webapp has changed phase to run at `prepare-package`, so that the npm build gets to finish before results are copied - the build and copy were running in the same phase so the result was indeterminate at best.

(if you don't understand what I'm talking about, read: https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html

To create a release: (if you have `geany` editor - otherwise substitute that)
```
touch rel-prepare.log rel-perform.log
geany rel-prepare.log rel-perform.log
mvn release:prepare -l rel-prepare.log -DautoVersionSubmodules=true -DreleaseVersion=5.7.8-GBI-M1 -Dtag=v5.7.8-GBI-M1 -DdevelopmentVersion=5.7.8-SNAPSHOT -T1
mvn release:perform -l rel-perform.log -T1
```

if you don't like the tag:
```
git push --delete origin v5.7.8-GBI-M1
git tag --delete v5.7.8-GBI-M1
```

### DIFF
--- a/viewer/pom.xml
+++ b/viewer/pom.xml
@@ -18,7 +18,9 @@
         <minify.outputdir>${project.name}-${project.version}/viewer-html</minify.outputdir>
         <viewerhtml.path>${basedir}/src/main/webapp/viewer-html</viewerhtml.path>
         <test.persistence.unit>viewer-config-hsqldb</test.persistence.unit>
-        <rest.api.outputdoc>${basedir}/../tailormap-components/projects/core/src/lib/shared/generated</rest.api.outputdoc>
+        <tailormap-components.source.dir>../tailormap-components</tailormap-components.source.dir>
+        <tailormap-components.dist.dir>${tailormap-components.source.dir}/dist/</tailormap-components.dist.dir>
+        <rest.api.outputdoc>${tailormap-components.source.dir}/projects/core/src/lib/shared/generated</rest.api.outputdoc>
     </properties>
     <dependencies>
         <dependency>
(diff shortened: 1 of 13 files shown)


viewer/pom.xml
```diff
@@ -224,19 +236,42 @@
...
+                        <goals>
+                            <goal>copy-resources</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>${project.build.directory}/${project.artifactId}-${project.version}/tailormap-components</outputDirectory>
+                            <resources>
+                                <resource>
+                                    <directory>${tailormap-components.dist.dir}/</directory>
+                                    <!--includes>
+                                        <include>... misschien *.js</include>
```
mprins, line 267: @mtoonen Not sure about this, there are some font files (material icons) in here as well that are probably needed

mprins:
I've created a release "5.7.8-GBI-M1" (tagged v5.7.8-GBI-M1) using the above commands, so if you want to re-create you need to delete the tags as shown above

The artifacts are available as:
- https://repo.b3p.nl/nexus/repository/releases/org/flamingo-mc/viewer-admin/5.7.8-GBI-M1/viewer-admin-5.7.8-GBI-M1.war
- https://repo.b3p.nl/nexus/repository/releases/org/flamingo-mc/viewer/5.7.8-GBI-M1/viewer-5.7.8-GBI-M1.war

mtoonen (approved):

PR #2055, opened 2020-10-14
GBIW-114: form cancel button doesnt work

mtoonen:


viewer-admin/src/main/webapp/WEB-INF/jsp/services/editform.jsp
```diff
@@ -49,7 +49,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                             <fmt:message key="viewer_admin.editattribute.3" var="editattribute3" />
                             <fmt:message key="viewer_admin.editattribute.4" var="editattribute4" />
                             <stripes:submit name="save" value="${editattribute3}"/>
-                            <stripes:reset name="cancel" class="extlikebutton" value="${editattribute4}"/>
+                            <stripes:submit name="cancel" class="extlikebutton" value="${editattribute4}"/>
```
mprins, line 52: So cancel is not supposed to reset the form, but submit the form to the [FormActionBean#cancel resolution](https://github.com/B3Partners/tailormap/blob/60acb684a981a75ce5017fc7dd9f88ce9c343318/viewer-admin/src/main/java/nl/b3p/viewer/admin/stripes/FormActionBean.java#L101-L103)
which then forwards to this edit form again (identical flow as #edit)

that seems convoluted, I guess for handling sideffects?...

mtoonen: People expect the cancel button in the iframe to cancel the action in the iframe

PR #2067, opened 2020-10-19
fixed circular dependency

mtoonen:
I think it is fixed, could you check if the functionality still works?


mtoonen:
Superseded by #2068

PR #2075, opened 2020-10-21
Feat/gbiw 177 proxy for rest calls

mtoonen:


mprins (changes requested):
I would strip protocol and hostname from the requests as typically this proxy only passes requests to localhost and not anywhere else.

so leaving out `window.location.origin` could do that, but will still leave the backend as an open proxy for any logged in users.


Also using an `ErrorResolution`  will break the contract as it returns html, could be fixed with something like this: 
https://github.com/B3Partners/tailormap/blob/7cedbf2a05770f454ad033af8e509cd198d068f1/viewer/src/main/java/nl/b3p/viewer/stripes/UserLayerActionBean.java#L95-L102

also code scanning flags quite a few issues in both java and typescript code

mprins (approved):

PR #2083, opened 2020-10-26
Fix performance of component configs

### LINKED ISSUE #2082: Performance issue: loading configuration of components
Loading a component configuration in a viewer with a _**large**_ layer tree takes ages (amplified when using oracle).

Cause:
appLayer.toJson()
---> addAttributes()

Only a small part of the components use the attributes (graph, simplefilter and cyclorama).

mtoonen:
- [x] Stop requesting attributes for all components
- [x] Fix Graph config
- [x] Fix simplefilter config
- [x] Fix cyclorama config

Fixes #2082

### DIFF
--- a/viewer-admin/src/main/java/nl/b3p/viewer/admin/stripes/LayoutManagerActionBean.java
+++ b/viewer-admin/src/main/java/nl/b3p/viewer/admin/stripes/LayoutManagerActionBean.java
@@ -214,7 +214,8 @@ public class LayoutManagerActionBean extends ApplicationActionBean {
         EntityManager em = Stripersist.getEntityManager();
 
         try {
-            appConfigJSON = application.toJSON(context.getRequest(),false, true, true, true,em, false).toString();
+            appConfigJSON = application.toJSON(context.getRequest(),false,
+                    true, false, false,em, false).toString();
         } catch(JSONException je) {
         }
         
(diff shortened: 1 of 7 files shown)


viewer-admin/src/main/java/nl/b3p/viewer/admin/stripes/LayoutManagerActionBean.java
```diff
@@ -213,12 +213,6 @@ public Resolution view() throws JSONException {
     public Resolution config() {
         EntityManager em = Stripersist.getEntityManager();
 
-        try {
-            appConfigJSON = application.toJSON(context.getRequest(),false,
-                    true, false, false,em, false).toString();
```
mprins: Why was it here? seems this is the only place it's used in this class and I can't find any other use... 
I think you could remove `appConfigJSON` from the class altogether

mprins (approved):
volgens mij kun je `appConfigJSON` uit de bean halen

viewer-admin/src/main/java/nl/b3p/viewer/admin/stripes/LayoutManagerActionBean.java
mtoonen: Done

PR #2089, opened 2020-10-28
Added columns attribute to filter out columsn for export

mtoonen:
Passing a list with column names will filter the exported attributes

### DIFF
--- a/viewer/src/main/java/nl/b3p/viewer/stripes/DownloadFeaturesActionBean.java
+++ b/viewer/src/main/java/nl/b3p/viewer/stripes/DownloadFeaturesActionBean.java
@@ -112,6 +112,9 @@ public class DownloadFeaturesActionBean extends LocalizableApplicationActionBean
     @Validate
     private String params;
 
+    @Validate
+    private List<String> columns = new ArrayList<>();
+
     private AuditMessageObject auditMessageObject;
 
     //<editor-fold defaultstate="collapsed" desc="getters and setters">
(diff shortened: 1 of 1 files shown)


viewer/src/main/java/nl/b3p/viewer/stripes/DownloadFeaturesActionBean.java
```diff
@@ -294,6 +308,23 @@ public void stream(HttpServletResponse response) throws Exception {
         }
     }
 
+    private List<ConfiguredAttribute> filterAttributes(List<ConfiguredAttribute> attributes) {
+        List<ConfiguredAttribute> newAttrs = new ArrayList<>();
+        if (!columns.isEmpty()) {
+            for (String column : columns) {
+                for (ConfiguredAttribute attr : attributes) {
```
mprins, line 315: chances are that `attributes` is always larger (and has larger objects) than `columns` so having the outer loop over `attributes` might be faster I think, as well as using a filtered stream.

mprins (approved):
OK, but could be improved

viewer/src/main/java/nl/b3p/viewer/stripes/DownloadFeaturesActionBean.java
mprins: also, after switching around, you can stop once `newAttrs.size()` reaches `columns.size()` as there won't be more to add

mtoonen: Funny, was actually optimising for that (see also: https://stackoverflow.com/questions/35710346/why-is-there-a-performance-difference-between-the-order-of-a-nested-loop). Because of the break on line 318, it would be faster to have the smaller array (columns) on the outside.

By inference, newAttrs.size() can never be larger than columns.size()

PR #2104, opened 2020-11-09
[mantis 15294] searchconfig loses its layers sometimes

mtoonen:
Old config was not retrieved and saved when it was not opened

Mantis-15294


mprins (approved):
OK I guess, 
so this is a regression of #1729

mtoonen:
Yeah regression is a nice way of describing my error :-p

PR #2111, opened 2020-11-12
Some code improvements

geertplaisier:


tailormap-components/projects/bridge/src/tailor-map.service.ts
```diff
@@ -43,7 +47,7 @@ export class TailorMapService {
     const mc = vc.mapComponent;
     const map = mc.getMap();
     map.addListener('ON_LAYER_VISIBILITY_CHANGED', (object, event) => {
-      this.layerVisibilityChanged$.next(event);
```
geertplaisier, line 46: These events indeed did not trigger change detection right away. Running inside the ngZone here removes the need to use ngZone where the Observable is subscribed to

tailormap-components/projects/bridge/typings/Components.d.ts
```diff
@@ -27,6 +27,9 @@ declare interface ViewerController {
   getAppLayerById: (appLayerId: number) => AppLayer;
   getAppLayer: (serviceId: number, layerName: string) => AppLayer;
 
-  getVisibleLayers: (castToStrings?: boolean) => number[] | string[];
+  getVisibleLayers(): number[];
```
geertplaisier, line 30: By using function overloads you can define the return type for different scenarios

tailormap-components/projects/core/src/lib/user-interface/add-feature/add-feature.component.ts
```diff
@@ -1,41 +1,47 @@
...
-    private ngZone: NgZone,
   ) {
-    this.tailorMapService.layerVisibilityChanged$.subscribe(value => {
-      this.ngZone.run(() => {
-        this.calculateVisibleLayers();
-      });
+    combineLatest([
+      this.tailorMapService.layerVisibilityChanged$,
+      this.formConfigRepo.formConfigs$,
+    ]).pipe(takeUntil(this.destroyed)).subscribe(() => {
```
geertplaisier, line 35: Always make sure to unsubscribe to observables

tailormap-components/projects/core/src/lib/user-interface/attributelist/attributelist-panel/attributelist-panel.component.html
```diff
@@ -33,7 +33,7 @@
   <ng-template #withLayers>
     <mat-tab-group #tabgroup animationDuration="0ms"
                    (selectedTabChange)="onSelectedTabChange();">
-      <mat-tab *ngFor="let layer of layers; let index = index" [label]="layer.name">
+      <mat-tab *ngFor="let layer of layers; let index = index; trackBy: trackByLayerId" [label]="layer.name">
```
geertplaisier, line 36: trackBy improves the performance a lot since Angular will not update once the result of this trackByLayerId method is equal

tailormap-components/projects/core/src/lib/user-interface/attributelist/layer.service.ts
```diff
@@ -14,27 +16,22 @@ export class LayerService {
...
   }
 
   public getAppId(): number {
     const vc = this.tailorMapService.getViewerController();
     return vc.app.id;
   }
 
-  public getLayers$(): Observable<Layer[]> {
-    this.loadLayers();
-    return of(this.layers);
```
geertplaisier, line 35: this returns an observable that is completed immediately. Rewrote that to a BehaviourSubject & Observable so subscriptions are actually updated once layers are updated.

tailormap-components/projects/bridge/typings/Components.d.ts
mtoonen: Well, now I feel stupid...

tailormap-components/projects/core/src/lib/user-interface/attributelist/attributelist-panel/attributelist-panel.component.html
mtoonen: Nice!

mtoonen (approved):

PR #2123, opened 2020-11-24
[Bug] Fix switching on layers via search doesn't work

mtoonen:
save layervisibility state when setting layervisibility, so switchin on layers via other components (such as search) works as well.

See mantis-15654


mprins (approved):
LGTM, but untested

PR #2163, opened 2020-12-11
GBIW-288: Show criteria description

geertplaisier:
Review eerst
- https://github.com/B3Partners/tailormap/pull/2161
- https://github.com/B3Partners/tailormap/pull/2162

Deze is gebaseerd op die 2 PR's


mtoonen (approved):
Number of features of resulting query not present, but something has to be left for next year

PR #2180, opened 2020-12-17
GBIW 204 Edit geometrie

mtoonen:
Also closes #2179


tailormap-components/projects/core/src/lib/shared/shared.module.ts
```diff
@@ -93,36 +93,37 @@ import {
...
-    MatDialogModule,
-    DialogCloseButtonComponent,
-    TreeComponent,
-    MatRadioModule,
-    MatProgressBarModule,
-    MatAutocompleteModule,
-    DragDropModule,
-    MatDatepickerModule,
-  ],
+        ReactiveFormsModule,
```
geertplaisier: oh no... why 4 spaces here? the rest is 2

tailormap-components/projects/core/src/lib/user-interface/geometry-confirm-buttons/geometry-confirm-buttons.component.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  constructor(private geometryConfirmService: GeometryConfirmService) {
+    this.position = {
+      left: 0,
+      top: 0,
+    };
+    this.geometryConfirmService.positionChanged$
+      .pipe(takeUntil(this.destroyed)).subscribe(value => {
+      this.position = {
+        left: value.left + 5,
```
geertplaisier: 5 is some margin & 50 is the height of the buttons? Maybe its nice to make variables for these values so that is clear what they mean

tailormap-components/projects/core/src/lib/user-interface/geometry-confirm-buttons/geometry-confirm-buttons.component.ts
```diff
@@ -0,0 +1,55 @@
...
+  private destroyed = new Subject();
+  public position: ScreenCoordinate;
+  public visible = false;
+
+  constructor(private geometryConfirmService: GeometryConfirmService) {
+    this.position = {
+      left: 0,
+      top: 0,
+    };
+    this.geometryConfirmService.positionChanged$
```
geertplaisier: You could save some lines of code here by not subscribing to 2 observable here but mapping to position and visiblity and using the async pipe in the template. That way you can remove the destroyed subject + OnInit & OnDestroy.

For example: 
`
this.position$ = combineLatest([
      this.geometryConfirmService.positionChanged$,
      this.geometryConfirmService.visibilityChanged$,
    ]).pipe(map(([ position, visible ]) => {
        return visible ? { left: position.left + MARGIN, top: position.top + BUTTON_HEIGHT } : null;
      }));
`

and in the template

`<div class="geometry-confirm" *ngIf="position$ | async as position" [style.left.px]="position.left [style.top.px]="position.top">`

tailormap-components/projects/core/src/lib/workflow/workflows/EditgeometryWorkflow.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    const dialogRef = this.dialog.open(FormComponent, {
+      id: this.FORMCOMPONENT_DIALOG_ID,
+      width: '1050px',
+      height: '800px',
+      disableClose: true,
+      data,
+    });
+    // tslint:disable-next-line: rxjs-no-ignored-subscription
+    dialogRef.afterClosed().subscribe(result => {
```
geertplaisier: You could also add takeUntil(this.destroyed) here? then you don't need the TS hint

tailormap-components/projects/core/src/lib/workflow/workflows/Workflow.helpers.ts
```diff
@@ -0,0 +1,48 @@
+import { GeoJSONGeometry } from 'wellknown';
```
geertplaisier: Nitpicking: the current convention is to use lowercase names for files (at least start lower case) + helper instead of helpers

tailormap-components/projects/core/src/lib/shared/shared.module.ts
mtoonen: Dunno why, reverted it

tailormap-components/projects/core/src/lib/user-interface/geometry-confirm-buttons/geometry-confirm-buttons.component.ts
mtoonen: Oeh, very nice.

tailormap-components/projects/core/src/lib/workflow/workflows/Workflow.helpers.ts
mtoonen: You mean the file "Workflow.helpers.ts" itself (line 1 is selected here, which has a lowercase as import)?
Didn't know this convention, but I'll use it

PR #2182, opened 2020-12-17
Feat/gbiw 278 position edit bar

mtoonen:
First review #2180 , this is based on that

@geertplaisier : is there any way to disable the animation of moving the dialog completely? Removing   transition: transform 0.5s; makes it only faster


tailormap-components/projects/core/src/lib/workflow/workflows/Workflow.helpers.ts
```diff
@@ -0,0 +1,48 @@
+import { GeoJSONGeometry } from 'wellknown';
+import { Geometry } from '../../shared/generated';
+import { Coordinate } from '../../user-interface/models';
+
+export class WorkflowHelpers {
+
+  public static findTopRight(geojson: GeoJSONGeometry | Geometry): Coordinate {
+    if (this.isGeneratedGeometry(geojson)) {
+      return this.findUpperRightCoordinate(geojson.coordinates[0] as number[][]);
```
geertplaisier: In general when accessing static methods the classname + method is used. I think both work but its more clear a static method is used instead of an instance method, so `WorkflowHelpers.findUpperRightCoordinate`

tailormap-components/projects/core/src/lib/workflow/workflows/Workflow.helpers.ts
```diff
@@ -0,0 +1,48 @@
+import { GeoJSONGeometry } from 'wellknown';
+import { Geometry } from '../../shared/generated';
+import { Coordinate } from '../../user-interface/models';
+
+export class WorkflowHelpers {
```
geertplaisier: This file was renamed in the other PR right? Remove here?

geertplaisier (changes requested):

PR #2185, opened 2020-12-21
Authorizations for forms created

mtoonen:
When no reader is selected for a form, it is for everybody

### DIFF
--- a/tailormap-components/projects/core/src/lib/shared/formconfig-repository/formconfig-repository.service.ts
+++ b/tailormap-components/projects/core/src/lib/shared/formconfig-repository/formconfig-repository.service.ts
@@ -1,5 +1,8 @@
 import { Injectable } from '@angular/core';
-import { HttpClient } from '@angular/common/http';
+import {
+  HttpClient,
+  HttpParams,
+} from '@angular/common/http';
 import {
   FormConfiguration,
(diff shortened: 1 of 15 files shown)


tailormap-components/projects/core/src/lib/shared/formconfig-repository/formconfig-repository.service.ts
```diff
@@ -32,7 +35,12 @@ export class FormconfigRepositoryService {
     private domainRepo: DomainRepositoryService,
     private tailorMap: TailorMapService,
   ) {
-    this.http.get<FormConfigurations>(this.tailorMap.getContextPath() + '/action/form')
+    let httpParams: HttpParams = new HttpParams();
```
geertplaisier: Small tip: if you create a new object which you are not going to change its not really needed to also specify the type.

Also, these HttpParams are chainable, for example, this is the same:

`const httpParams = new HttpParams().set('application', '' + this.tailorMap.getViewerController().app.id);`

mprins: ja... wat Geert zegt :wink:

mprins (approved):
LGTM

PR #2214, opened 2021-01-11
Fix saving new geometries

mtoonen:
reset objectguid to feature so it isn't set to -1. Fixes error when saving: -1 objectguid is a new object. When an existing object with related features is editted, saving it as a new object would remove the related features (which is not good)

### DIFF
--- a/tailormap-components/projects/core/src/lib/workflow/workflows/EditgeometryWorkflow.spec.ts
+++ b/tailormap-components/projects/core/src/lib/workflow/workflows/EditgeometryWorkflow.spec.ts
@@ -15,14 +15,14 @@ import { mockBoom, vectorLayerMock } from '../../shared/tests/test-data';
 import { OLFeature } from '../../../../../bridge/typings';
 import { Coordinate } from '../../user-interface/models';
 import { createMockDialogProvider } from './mocks/workflow.mock';
+import { getTailorMapServiceMockProvider } from '../../../../../bridge/src/tailor-map.service.mock';
 
 describe('EditgeometryWorkflow', () => {
   let workflow : EditgeometryWorkflow;
   let factory: SpectatorService<WorkflowFactoryService>;
 
   const vectorLayer = vectorLayerMock({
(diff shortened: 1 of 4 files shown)


tailormap-components/projects/core/src/lib/workflow/workflows/EditgeometryWorkflow.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const vectorLayer = vectorLayerMock({
+    getActiveFeature(): OLFeature {
+      return {color: '', config: {wktgeom: '', id: 'string', attributes: {}}}
+    },
+  });
+
+  const geometryConfirmService = createSpyObject(GeometryConfirmService, {
+    open(coord: Coordinate): Observable<boolean> {
+      return new BehaviorSubject<boolean>(true).asObservable();
```
geertplaisier, line 31: This could also be simply `of(true)` I think?

tailormap-components/projects/core/src/lib/workflow/workflows/EditgeometryWorkflow.ts
```diff
@@ -46,11 +46,11 @@ export class EditgeometryWorkflow extends Workflow {
...
   private openForm(geom: GeoJSONGeometry) {
-
-    const objecttype = this.event.feature.objecttype;
+    const feature = this.event.feature;
+    const objecttype = feature.objecttype;
     const feat = this.featureInitializerService.create(objecttype,
-      {...this.event.feature, geometrie: geom  });
-
+      {...feature, geometrie: geom  });
+    feat.objectGuid = feature.objectGuid;
```
geertplaisier, line 53: If you expect this to happen more in the future, maybe its nice to add as argument to `featureInitializerService.create` (`public create(type: string, params: any, objectGuid = FeatureInitializerService.STUB_OBJECT_GUID_NEW_OBJECT): Feature {`)

geertplaisier (approved):

PR #2286, opened 2021-01-27
CI/CD only triggers workflows for its intended jobs

mtoonen:
Only run angular jobs on *.ts files, and maven jobs on *.java and *.xml.

Started implementing windows workflow. I think this could replace appveyor?


mprins:
> I think this could replace appveyor?

Appveyor runs integration tests with MS SQL Server on windows, this PR doesn't provide that

.github/workflows/angular-components.yml
```diff
@@ -1,6 +1,12 @@
 name: Angular components CI
 
-on: [push, pull_request]
+on:
+  push:
+    paths:
+    - '**.ts'
+  pull_request:
+    paths:
+    - '**.ts'
```
geertplaisier: Also add *.css and *.html here since those are part of Angular code as well

geertplaisier (changes requested):

.github/workflows/angular-components-deployment.yml
```diff
@@ -7,7 +7,9 @@ on:
 #      - feat/GBIW-142_continuous_deployment
   pull_request:
     branches:
-      - feat/gbi-component
+      - master
+  paths:
+    - '**.ts'
```
geertplaisier: Also add *.css and *.html here since those are part of Angular code as well

geertplaisier (changes requested):

mprins:
Not in favour of this; it will make it hard to require build succes criteria for a PR as we do now, also it doen't cover changes to eg. GH workflows and regular Javascript (which is build using Maven)

geertplaisier:
The problem now is that often we (or is it just me?) do not really look at the CI results since the results are not always correct. Especially if I only change something in Angular code, why would we run all the Java tests / builds as well. In that case I have to scroll through the list to find the relevant checks and see if those are green.

If we would only run the jobs which are impacted by the change it makes it a lot easier to see whether your changes actually broke something.

Maybe we can change the checks introduced in this PR by exluding tailormap-components on the Java checks/builds? So when you only make changes to the Angular code it does not run the rest?

mtoonen:
> Not in favour of this; it will make it hard to require build succes criteria for a PR as we do now, also it doen't cover changes to eg. GH workflows and regular Javascript (which is build using Maven)

Some additions must be made for workflows and regular javascript. 

But the goal remains te same: speeding up buildtime for angular/ts files: the success of a complete build is independent of angular as we test them extensively in the workflows. But java build slow down development, and make quick PR's with a few lines in TS impossible.

So, if we exclude java builds from ts edits, that would still work no? Can we stop triggering appveyor and jenkings for ts edits?

.github/workflows/windows-maven.yaml
```diff
@@ -0,0 +1,30 @@
...
+        restore-keys: |
+          ${{ runner.os }}-maven-
+    - name: Set up JDK
+      uses: actions/setup-java@v1
+      with:
+        java-version: ${{ matrix.java }}
+    - name: Build with Maven
+      run: |
+        mvn -B test -fae
+        mvn -B verify -fae -e
```
mprins: ik denk niet dat dit doet wat je denkt, voor integratie tests is een profile nodig en extra opties

mprins (commented):
ik zie het nut niet van een windows build op GH.
Geert vindt dat ie nu al door een lange lisjt heen moet scrollen, zo komt er nog 1 bij

.github/workflows/windows-maven.yaml
mtoonen: Het was vooral een uitprobeersel, om te kijken of ik mssql hier werkend kon krijgen, zodat we appveyor niet meer nodig hadden, maar dat was vanmorgen een beetje gestrand, dus die kan weg

geertplaisier (approved):

.github/workflows/windows-maven.yaml
mprins: Running Microsoft SQL Server 2019 Developer Edition in a Docker Windows Container

https://www.simteq.com/2020/05/14/running-microsoft-sql-server-2019-developer-edition-in-a-docker-windows-container/

mprins:
@mtoonen  Er zijn nu dus geen volledige builds meer op GH als er "klassieke" javascript code wordt aangepast; dan moeten we dus afwachten wat AppVeyor en Jenkins er van vinden en die zijn niet het meest stabiel, noch voor iedereen toegankelijk

zie bijv #2294

PR #2341, opened 2021-02-17
change style of buttons in formcopy component

mtoonen:


tailormap-components/projects/core/src/lib/feature-form/form-copy/form-copy.component.css
```diff
@@ -23,6 +23,15 @@
...
+.mat-primarylight {
+  background-color: rgba(98, 54, 255, 0.1);
+  color: #6236ff;
+}
+
 .copy-side-panel {
   width: 50%;
 }
+
+.actions button {
```
geertplaisier: Maybe change this to

```
.actions button + button {
   margin-left: 8px;
}
```

This way you don't have margin on the last button which you dont need

tailormap-components/projects/core/src/lib/feature-form/form/form.component.css
```diff
@@ -45,7 +45,7 @@
 
 .bottom-bar button {
   float: right;
-  margin-right: 15px;
+  margin-right: 8px;
```
geertplaisier, line 48: See other comment

geertplaisier (changes requested):

PR #2352, opened 2021-02-18
15882: retrieve features for the form tree from the state

mtoonen:
Fixes mantis-15882

### DIFF
--- a/tailormap-components/projects/core/src/lib/feature-form/form-tree/form-tree.component.ts
+++ b/tailormap-components/projects/core/src/lib/feature-form/form-tree/form-tree.component.ts
@@ -5,7 +5,7 @@ import { FormTreeHelpers } from './form-tree-helpers';
 import { Store } from '@ngrx/store';
 import { FormState } from '../state/form.state';
 import * as FormActions from '../state/form.actions';
-import { selectFormConfigs } from '../state/form.selectors';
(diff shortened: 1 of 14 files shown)


tailormap-components/projects/core/src/lib/feature-form/form-tree/form-tree.component.ts
```diff
@@ -72,9 +71,14 @@ export class FormTreeComponent implements OnInit, OnChanges, OnDestroy {
       },
       this.hasCheckboxes,
     );
-    if (this.features && this.features.length > 0) {
-      this.createTree(this.features);
-    }
+
+    this.store$.select(selectOpenFeatureForm).pipe(takeUntil(this.destroyed)).subscribe((features) => {
```
geertplaisier: I see the name selectOpenFeatureForm was already there but its a bit misleading, I thought it was meant to show if the form is open or not but its actually selecting features? Maybe consider renaming?

geertplaisier (approved):

tailormap-components/projects/core/src/lib/feature-form/form-tree/form-tree.component.ts
mtoonen: Only changed the reducer... will rename selector as well :)

PR #2354, opened 2021-02-18
15882: Creating new object fixed

mtoonen:
mantis-15882


tailormap-components/projects/core/src/lib/feature-form/form/form.component.ts
```diff
@@ -105,16 +110,26 @@ export class FormComponent implements OnDestroy, OnChanges, OnInit {
...
   }
 
+  public setFormEditting(editting){
+    this.store$.dispatch(FormActions.setFormEditting({editting: editting}));
+  }
+
   public newItem(evt) {
     const type = LayerUtils.sanitizeLayername(evt.srcElement.id);
     this.store$.select(selectFormConfigForFeatureType, type)
       .pipe(takeUntil(this.destroyed))
```
geertplaisier: This does not seem right... Creating a new item is an action by a user so this could be triggered multiple times. Each time you have a subscription which creates a new object. I think you need take(1) instead of takeUntil here

tailormap-components/projects/core/src/lib/feature-form/form/form.component.ts
```diff
@@ -105,16 +110,26 @@ export class FormComponent implements OnDestroy, OnChanges, OnInit {
...
-          this.initForm();
+        const objecttype = FormHelpers.capitalize(type);
+        const newFeature = this.featureInitializerService.create(objecttype, {
+          id: null,
+          clazz: type,
+          isRelated: true,
+          objecttype,
+          children: null,
         });
+        newFeature[formConfig.treeNodeColumn] = 'Nieuwe ' + formConfig.name;
```
geertplaisier: In Typescript you can now do this: 

```
const newFeature = this.featureInitializerService.create(objecttype, {
      id: null,
      clazz: type,
      isRelated: true,
      objecttype,
      children: null,
      [formConfig.treeNodeColumn]: 'Nieuwe ' + formConfig.name,
    });
```

And even better, replace the `'Nieuwe ' + formConfig.name` by \`Nieuwe ${formConfig.name}\`

tailormap-components/projects/core/src/lib/feature-form/state/form.reducer.ts
```diff
@@ -15,7 +15,21 @@ const onSetFeature = (state: FormState, payload : ReturnType<typeof FormActions.
   feature: payload.feature,
 });
 
-const onSaveFeature = (state: FormState, payload:  ReturnType<typeof FormActions.setSaveFeatures>) : FormState => ({
+const onSetNewFeature = (state: FormState, payload: ReturnType<typeof FormActions.setNewFeature>): FormState => ({
+  ...state,
+  features: [
+    {
+      ...state.features[0],
```
geertplaisier: Is this correct? Are you *always* updating the first one? Then its ok, but still it could be better to lookup the right feature first based on id or something

geertplaisier (changes requested):

mvdstruijk (approved):

geertplaisier (approved):

PR #2362, opened 2021-02-19
Move FormConfigs to Application state

geertplaisier:
Fixes
mantis-15906
mantis-15873


tailormap-components/projects/core/src/lib/feature-form/form-copy/form-copy.component.ts
```diff
@@ -162,12 +174,12 @@ export class FormCopyComponent implements OnInit, OnDestroy {
   }
 
   public isFieldChecked(event: any) {
-    const fieldsToCopy = this.formCopyService.featuresToCopy.get(this.originalFeature['objectGuid']);
+    const fieldsToCopy = this.formCopyService.featuresToCopy.get(this.originalFeature.objectGuid);
     return fieldsToCopy.has(event);
   }
 
   public isEverythingChecked(tab: string): boolean {
-    const fieldsToCopy = this.formCopyService.featuresToCopy.get(this.originalFeature['objectGuid']);
```
mtoonen, line 170: not sure if this will pass the test of time: it is now very tightly coupled with GBI features (they contain an objectGuid). For future purposes it might be better to create a field (lets call it __fid) where we store the primary key and use that. For now, I think it will do

tailormap-components/projects/core/src/lib/shared/attribute-service/attribute.service.ts
```diff
@@ -38,10 +38,6 @@ export class AttributeService {
    */
   public features$(params: AttributeListParameters): Observable<AttributeListResponse> {
     let httpParams: HttpParams = new HttpParams();
-    httpParams = httpParams.set('limit', '10');
-    httpParams = httpParams.set('page', '1');
-    httpParams = httpParams.set('start', '1');
-
```
mtoonen, line 44: No defaults?

mtoonen (approved):

PR #2395, opened 2021-02-24
15929: add new items to tree

mtoonen:
hopefully the last for mantis-15929


tailormap-components/projects/core/src/lib/feature-form/state/form.state-helpers.ts
```diff
@@ -31,3 +23,16 @@ export const updateFeatureInArray = (features: Feature[], newFeature: Feature):
       children: feature.children ? updateFeatureInArray(feature.children, newFeature) : null,
     }));
 };
+
+export const addFeatureToParent = (features: Feature[], newFeature: Feature, parentId:string) : Feature[] =>{
+  const idx = features.findIndex(feature => feature.objectGuid === parentId);
+
+  return (idx !== -1 ?
+      [...features.slice(0, idx),{...features[idx],children: [...features[idx].children, newFeature]},...features.slice(idx + 1)]
```
geertplaisier: I think this line is correct but a bit hard to read. Maybe add some breaks between the different parts?

tailormap-components/projects/core/src/lib/feature-form/state/form.state-helpers.ts
```diff
@@ -31,3 +23,16 @@ export const updateFeatureInArray = (features: Feature[], newFeature: Feature):
       children: feature.children ? updateFeatureInArray(feature.children, newFeature) : null,
     }));
 };
+
+export const addFeatureToParent = (features: Feature[], newFeature: Feature, parentId:string) : Feature[] =>{
+  const idx = features.findIndex(feature => feature.objectGuid === parentId);
+
+  return (idx !== -1 ?
```
geertplaisier: Since its already a somewhat complicated statement you could also do an early return:

```
if (idx === -1) {
    return features;
}
return {
   ...features.slice etc
```

geertplaisier (changes requested):

mvdstruijk (approved):

PR #2397, opened 2021-02-24
15892: Save feature after geometry edit

geertplaisier:


tailormap-components/projects/core/src/lib/workflow/workflows/EditgeometryWorkflow.ts
```diff
@@ -41,7 +41,7 @@ export class EditgeometryWorkflow extends Workflow {
       this.vectorLayer.readGeoJSON(geom);
```
mtoonen: Isn't the editgeometryworkflow obsolete now that the edit-feature-geometry.service.ts exists?

viewer/src/main/webapp/viewer-html/common/viewercontroller/ViewerController.js
```diff
@@ -469,6 +469,25 @@ Ext.define("viewer.viewercontroller.ViewerController", {
...
+                fillcolor: "FF0000",
+                fillopacity: 50,
+                strokecolor: "FF0000",
+                strokeopacity: 50
+            }
+        });
+        this.mapComponent.getMap().addLayer(layer);
+        return layer;
+    },
+
```
mtoonen, line 490: Hmm, not entirely happy with this, but as the ext side will be removed, lets not split hair ;-)
Probably did this to create better readable code in angular, so I think it's acceptable

mtoonen (approved):
Looks good, but I think editgeometryworkflows.ts can be removed, no?

viewer/src/main/webapp/viewer-html/common/viewercontroller/ViewerController.js
geertplaisier: Yep, I wanted to keep the specifics of how to create a layer in the 'old' code so edit-geometry-services does not need to know how to create one.

tailormap-components/projects/core/src/lib/workflow/workflows/EditgeometryWorkflow.ts
geertplaisier: Removed

PR #2398, opened 2021-02-24
15886: Allow feature selection before editing

geertplaisier:


mtoonen:
Hmm, I would expect a cancel/close button.
When clicking on a location where lots of items are located, this list becomes quite long. Is it a good idea to group them with accordions or something like that?

mtoonen (changes requested):
Hmm, I would expect a cancel/close button.
When clicking on a location where lots of items are located, this list becomes quite long. Is it a good idea to group them with accordions or something like that?

geertplaisier:
> Hmm, I would expect a cancel/close button.
> When clicking on a location where lots of items are located, this list becomes quite long. Is it a good idea to group them with accordions or something like that?

Cancel button added. Maybe an accordion helps but if there are lots of items, cancelling and zooming works better to get the item you want to edit I think? Especially since the name / label of the item is not always very descriptive.

PR #2399, opened 2021-02-24
15875: Apply Attribute List configuration

geertplaisier:


tailormap-components/projects/core/src/lib/user-interface/attributelist/attribute-list-button/attribute-list-button.component.ts
```diff
@@ -13,9 +13,15 @@ import { take } from 'rxjs/operators';
 })
 export class AttributeListButtonComponent implements OnInit {
 
+  public tooltip: string;
+
   @Input()
-  public set attributeListConfig(config: AttributeListConfig) {
+  public set attributeListConfig(configJsonString: string) {
+    const config: AttributeListConfig = JSON.parse(configJsonString);
     this.store$.dispatch(setAttributeListConfig({ config: { ...config } }));
```
mtoonen: Is the spread-operator necessary? I thought JSON.parse already cloned it

tailormap-components/projects/core/src/lib/user-interface/attributelist/attribute-list-panel/attribute-list-panel.component.ts
```diff
@@ -57,6 +58,8 @@ export class AttributeListPanelComponent implements OnInit, AfterViewInit, Layou
       .subscribe(tabs => {
         this.tabs = tabs;
       });
+    this.title$ = this.store$.select(selectAttributeListConfig)
```
mtoonen, line 61: Why do this differently than the tooltip in attribute-list-button.component.ts? Title isn't going to change (I think?)?

I don't care which one is used, but I merely noticed the difference

mtoonen (approved):

tailormap-components/projects/core/src/lib/user-interface/attributelist/attribute-list-panel/attribute-list-panel.component.ts
geertplaisier: In the button, the config is passed through an @Input by Ext. Then its added to the state. Since we already have the config there I did not read it from the state as well.

tailormap-components/projects/core/src/lib/user-interface/attributelist/attribute-list-button/attribute-list-button.component.ts
geertplaisier: Yes, good catch, removed.

PR #2436, opened 2021-03-03
15884: set scale higher when its lower than 1

mvdstruijk:


viewer/src/main/webapp/viewer-html/components/GBI.js
```diff
@@ -126,7 +126,9 @@ Ext.define("viewer.components.GBI", {
         var x = parseInt(coords.x);
         var y = parseInt(coords.y);
         var scale = this.config.viewerController.mapComponent.getMap().getResolution() * 4;
-
+        if (scale <= 1) {
```
geertplaisier: Why not check if scale <= 1.2, right now when scale is 1.1 it is less then when scale would be 1

Could also be a nice one-liner:
`var scale = Math.max(this.config.viewerController.mapComponent.getMap().getResolution() * 4, 1.2)`

geertplaisier (changes requested):

geertplaisier (approved):

PR #2449, opened 2021-03-04
15935: Fixed always active bug for simple drawing + fixed some layout issues

geertplaisier:
mantis-15935

Drawing tool was always active when in Simple drawing mode. Also fixed a couple of layout issues. Smaller windows sizes can now be set so you don't have a huge popup in simple drawing mode.

Removed left/right margins from layout options for Drawing component. Adding the drawing component there does not work very well, for example, the Close button is still visible and gives an error on clicking.

### DIFF
--- a/viewer/src/main/webapp/viewer-html/common/ScreenPopup.js
+++ b/viewer/src/main/webapp/viewer-html/common/ScreenPopup.js
@@ -53,7 +53,8 @@ Ext.define ("viewer.components.ScreenPopup",{
             changeableSize:true,
             items:null,
             position : 'center',
-            useExtLayout: false
+            useExtLayout: false,
+            usePopupScrolling: true
         }
     },
     component: null,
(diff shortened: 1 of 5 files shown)


mprins (approved):
lijkt me een verbetering :+1:

viewer/src/main/webapp/viewer-html/common/ScreenPopup.js
```diff
@@ -77,13 +78,13 @@ Ext.define ("viewer.components.ScreenPopup",{
             layout: 'fit',
             modal: false,
             renderTo: Ext.getBody(),
-            scrollable: true,
+            scrollable: this.config.details.usePopupScrolling,
             constrainHeader: true,
             iconCls: this.config.iconCls || "",
             bodyStyle: {},
-            cls: "screen-popup"
+            cls: "screen-popup",
```
mprins, line 85: de slepende komma is een probleem voor de javascript compiler in 5.6.x, dus die zal ik oplossen

PR #2471, opened 2021-03-12
Add hyperlink form field type in passport

geertplaisier:
Also adjusted design of form component to match UI design


mtoonen (approved):
Ok, but I would've made a separate class for the link field-type.  Can be refactored later if needed

PR #2506, opened 2021-03-19
15993: Refactor statistics by moving stats to the state + refresh after changing filters

geertplaisier:
Resolves mantis-15993


mtoonen (approved):
lgtm. Hard to see what was changed, but it works again

PR #2509, opened 2021-03-19
Backport/2472 fromconfigtoarray

mtoonen:


mvdstruijk:
This PR has TSLint errors, weird that this is not detected on the PR for the master.
@geertplaisier Do you have any idea? Maybe it has something todo with the switch from ESLint to TSLint?

geertplaisier:
Yes, I think you are right, in 5.8.x branch TSLint is used and in master ESLint is used. There are some minor differences between the two so I guess thats the reason for the errors here.

geertplaisier:
Although, TSLint is right here, store$ in DomainRepositoryService is indeed not used.

geertplaisier:
Fixed the lint error in this PR. For master I propose to impose stricter rules so this error won't happen again, see: https://github.com/B3Partners/tailormap/pull/2510

geertplaisier (approved):

PR #2510, opened 2021-03-19
Implement stricter linting: No unused vars / locals

geertplaisier:


tailormap-components/projects/bridge/tsconfig.spec.json
```diff
@@ -5,7 +5,8 @@
     "types": [
       "jasmine",
       "node"
-    ]
+    ],
+    "noUnusedLocals": false
```
mtoonen, line 9: Shouldn't this be true?
default is false, and we want checking for unused local variables don't we?

mtoonen (changes requested):

tailormap-components/projects/bridge/tsconfig.spec.json
geertplaisier: Well this file is used when testing, so for checking the spec.ts files. In that case I think it can make sense to have unused variables and is at least not a bad thing since its not production facing code. For example, in a lot of tests we have a `let store: MockStore` which we initialize but never use (yet). When adding tests its nice that its already there.

mtoonen (approved):

PR #2547, opened 2021-04-01
Improve performance by wrapping create-layer outside ngzone

geertplaisier:


matthijsln:
It works but we could also wrap/decorate the viewerController so that it runs outside the zone, but lot more work.

mvdstruijk (approved):

PR #2580, opened 2021-04-07
15958: only disable save button when the first config is solr type

mvdstruijk:
[fixes-mantis-15958](https://mantis.b3p.nl/view.php?id=15985)

In relation with:
- https://mantis.b3p.nl/view.php?id=15294
- https://github.com/B3Partners/tailormap/issues/2013

### DIFF
--- a/viewer/src/main/webapp/viewer-html/components/Search-config.js
+++ b/viewer/src/main/webapp/viewer-html/components/Search-config.js
@@ -36,7 +36,13 @@ Ext.define("viewer.components.SearchConfiguration",{
         if (configObject === null){
             configObject = {};
         }
-        configObject.willLoadLayers = true;
+        if(configObject.searchconfigs) {
+            if(configObject.searchconfigs.length > 0) {
+                if (configObject.searchconfigs[0].type === 'solr') {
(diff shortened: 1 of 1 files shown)


viewer/src/main/webapp/viewer-html/components/Search-config.js
```diff
@@ -36,7 +36,13 @@ Ext.define("viewer.components.SearchConfiguration",{
         if (configObject === null){
             configObject = {};
         }
-        configObject.willLoadLayers = true;
+        if(configObject.searchconfigs) {
```
geertplaisier, line 39: Could be 1 if-statement

```
if(configObject.searchconfigs
    && configObject.searchconfigs.length > 0
    && configObject.searchconfigs[0].type === 'solr'
) {
    configObject.willLoadLayers = true;
}
```

geertplaisier (approved):

viewer/src/main/webapp/viewer-html/components/Search-config.js
geertplaisier: Ooops, does not work when configObject.searchConfigs is undefined

PR #2606, opened 2021-04-13
set restriction to attribute, and load infocomponents after all compo…

mtoonen:
…nents are finished loading

See MANTIS-15931 : featurereport could not load layers due to empty restriction. Also fixed a timingissue for loading the info components


mprins (approved):
LTGM

PR #2611, opened 2021-04-14
16502: copy mainFeatures objects from parentFeature

mvdstruijk:
[fixes-mantis-16502](https://mantis.b3p.nl/view.php?id=16052)


mtoonen (approved):
LGTM

PR #2620, opened 2021-04-15
check if the value was indeed in the list of options. if not, use the…

mtoonen:
… entered value

mantis-16061

TM-43


geertplaisier (approved):

geertplaisier (changes requested):
Maybe better to check for `typeof compVal !== 'undefined' ? compVal.val : featureAttribute.value`

Right now, if compVal would be 0 or an empty string for example featureAttribute.value would be used which could be wrong

mvdstruijk (approved):

PR #2643, opened 2021-04-19
Tm-32: added option to filter on NULL values

mvdstruijk:
https://b3partners.atlassian.net/jira/software/projects/TM/boards/14?selectedIssue=TM-32


tailormap-components/projects/core/src/lib/shared/helpers/attribute-filter.helper.ts
```diff
@@ -76,6 +82,9 @@ export class AttributeFilterHelper {
     if (filter.condition === 'ENDS_WITH') {
       query.push(AttributeTypeHelper.getExpression(`%${value}`, AttributeTypeEnum.STRING));
     }
+    if (filter.condition === 'NULL') {
```
geertplaisier: This is never reached?

tailormap-components/projects/core/src/lib/shared/helpers/attribute-filter.helper.ts
```diff
@@ -63,6 +64,11 @@ export class AttributeFilterHelper {
     if (filter.condition === 'NOT_LIKE') {
       query.push('NOT');
     }
+    if (filter.condition === 'NULL') {
+      query.push('IS');
```
geertplaisier: query.push('IS NULL'); is the same and maybe a bit more readable

geertplaisier (changes requested):

tailormap-components/projects/core/src/lib/shared/helpers/attribute-filter.helper.ts
mvdstruijk: Yes i thought the same but i was not sure if that's a problem because you have 2 indexes in the array instead of 3

geertplaisier: no, the array is just for building up a string and is joined with a space later on so does not matter how many items it contains

tailormap-components/projects/core/src/lib/shared/helpers/attribute-filter.helper.ts
mvdstruijk: Good catch, old line -_-

geertplaisier (approved):

PR #2652, opened 2021-04-21
Improve table performance

geertplaisier:


mtoonen (approved):
Paging with the devtools enabled is still quite slow, but it is less noticeable

PR #2658, opened 2021-04-22
add parentheses around filter to fix filter retrieval

mtoonen:
Fixes TM-21


mvdstruijk:
Not sure but is it not better to prevent l.isUserLayer() to be null, so it always is a boolean with default false?

mtoonen:
> Not sure but is it not better to prevent l.isUserLayer() to be null, so it always is a boolean with default false?

Hmm, I think it would have been better, but that is a larger change, as it would require some update script for pre existing layers. Maybe we should get rid of this later

mvdstruijk (approved):

PR #2662, opened 2021-04-22
TM-11: use css variable for multiple columns in passport

mvdstruijk:
Fixes TM-11

### DIFF
--- a/tailormap-components/projects/core/src/lib/feature-form/form-creator/form-creator.component.css
+++ b/tailormap-components/projects/core/src/lib/feature-form/form-creator/form-creator.component.css
@@ -1,28 +1,30 @@
 :host,
 form,
-mat-tab-group {
+mat-tab-group,
+.column-container {
   height: 100%;
 }
 
-.tab {
-  float: left;
-  width: calc(100 / var(--overlay-panel-columns) * 1%);
-}
-
-.is-editing .tab {
-  padding-bottom: 16px;
+.column-container {
+  display: flex;
 }
 
 .column {
-  display: flex;
-  flex-direction: column;
-  height: 77px;
+  flex: 1 0;
 }
 
-.column + .column {
+.is-editing .column {
+  padding-bottom: 16px;
+}
+
+.row {
+  min-height: 77px;
+}
+
+.row + .row {
   border-top: 1px solid rgba(216, 216, 216, 0.62);
 }
 
-.is-editing .column + .column {
+.is-editing .row + .row {
   border-top: 0 none;
 }
--- a/tailormap-components/projects/core/src/lib/feature-form/form-creator/form-creator.component.css
+++ b/tailormap-components/projects/core/src/lib/feature-form/form-creator/form-creator.component.css
@@ -4,6 +4,10 @@ mat-tab-group {
   height: 100%;
 }
 
+.tab {
+  float: left;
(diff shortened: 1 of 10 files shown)


mtoonen (approved):
Seems to work :+1:

tailormap-components/projects/core/src/lib/feature-form/form-creator/form-creator.component.css
```diff
@@ -4,13 +4,19 @@ mat-tab-group {
   height: 100%;
 }
 
+.tab {
+  float: left;
```
geertplaisier: Maybe better to use a display: flex parent & set flex: 1 on the tabs. Floating can break your layout quite easily and then you also don't need to set a width

tailormap-components/projects/core/src/lib/feature-form/form-creator/form-creator.component.css
```diff
@@ -4,13 +4,19 @@ mat-tab-group {
...
+}
+
 .is-editing .tab {
   padding-bottom: 16px;
 }
 
 .column {
   display: flex;
   flex-direction: column;
+  height: 77px;
```
geertplaisier: Why is this needed?

tailormap-components/projects/core/src/lib/feature-form/form-creator/form-creator.component.html
```diff
@@ -1,8 +1,8 @@
 <form [formGroup]="formgroep">
   <mat-tab-group class="tab-group" [class.is-editing]="editing$ | async">
     <mat-tab *ngFor="let tab of tabbedConfig; trackBy: trackByTabId" label="{{formConfig.tabConfig[tab.tabId]}}">
-      <div class="tab">
-        <div *ngFor="let attr of tab.attributes" class="column">
+      <div *ngFor="let column of getColumns()" class="tab">
```
geertplaisier: This is not really efficient and happens on every CD

tailormap-components/projects/core/src/lib/feature-form/form-creator/form-creator.component.ts
```diff
@@ -66,6 +66,8 @@ export class FormCreatorComponent implements OnChanges, OnDestroy, AfterViewInit
   private destroyed = new Subject();
 
   public ngOnChanges() {
+    document.documentElement.style.setProperty('--overlay-panel-columns', Math.max.apply(Math,this.formConfig.fields.map(field=> field.column)));
```
geertplaisier: I think its more efficient to create a Column model or something. So you only have to create the columns once and you could also add the attributes to this column model so you don't need the getAttributes method below.

tailormap-components/projects/core/src/styles/styles.css
```diff
@@ -68,7 +70,7 @@
   right: 10px;
   top: 10px;
   bottom: 10px;
-  width: 300px;
+  width: calc(var(--overlay-panel-columns) * var(--overlay-panel-width));
```
geertplaisier: Wouldn't do this in the styles.css because they are meant to be global and this is quite specific

geertplaisier (changes requested):

geertplaisier (approved):

PR #2694, opened 2021-05-20
added primaryKeyAttribute to simplefeaturetype. This is filled when u…

mtoonen:
…pdating/creating a JDBC source

Fixes TM-146

@matthijsln  @mprins It seems only JDBCDataStore supports exposing the PK column, however, there is a checkbox in geoserver to expose them for WFS sources. Do you know how to retrieve that information via geotools?

### DIFF
--- a/tailormap-components/projects/core/src/lib/shared/attribute-service/attribute-models.ts
+++ b/tailormap-components/projects/core/src/lib/shared/attribute-service/attribute-models.ts
@@ -51,6 +51,7 @@ export interface AttributeMetadataResponse {
   featureType: number;
   featureTypeName: string;
   geometryAttribute: string;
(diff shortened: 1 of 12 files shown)


mprins:
>  Do you know how to retrieve that information via geotools?

depends on the output format of the WFS... In GML there is an attribute `gml:id` on a Feature which should have the PK value. I think geotools should be able to give the value of the `fid` (which may or may not be an actual primary key somewhere - that depends on the datasource of the WFS)

mprins:
I don't think the FeaturetypeDescription lists a primary key; it will just list the constraints for an attrtibute, eg. `unique`, `not null`, so you could guess using a `fid` value which columns compose the PK for a featuretype

PR #2715, opened 2021-06-03
Use @tailormap/core-components and @tailormap/shared

geertplaisier:
Use @tailormap/core-components and @tailormap/shared packages
Removed attribute-list, tree and some helpers, those are now used from @tailormap/core-components and @tailormap/shared


tailormap-components/projects/core/src/lib/analysis/state/analysis.effects.ts
```diff
@@ -32,7 +28,7 @@ export class AnalysisEffects {
   ));
 
   constructor(
-    private store$: Store<AnalysisState>,
+    private store$: Store,
```
mtoonen, line 31: Why lose the typing? Isn't it better to define which states the component uses?

mtoonen (approved):
Works on my machine :+1: 

Very nice!

tailormap-components/projects/core/src/lib/analysis/state/analysis.effects.ts
geertplaisier: Because the masters say so: https://github.com/timdeschryver/eslint-plugin-ngrx/blob/main/docs/rules/no-typed-global-store.md

Not 100% sure myself yet, it could help identify / think about which feature state you want to use, even though it means nothing because global store is always injected

PR #2747, opened 2021-06-15
TM-10 added checkboxes and date types for form-field

mvdstruijk:
Added checkboxes and date types
TM-10

To configure a datepickerfield:
Set type to: date
and add dateFormat to the formconfig:
Example: {
    "key" : "inspectiedatum",
    "type" : "date",
    "column" : 1,
    "tab" : 2,
    "label" : "Inspectiedatum",
    "mandatory" : false,
    "linkedList" : 0,
    "dateFormat": "YYYY-MM-DD"
  }

To configure a checkbox:
Set type to: checkbox
Add the value for True and False as it is in the DB to the formconfig. For example if true = "Y" and false = "N".
Example:
{
    "key" : "m4_structurelewaarde",
    "type" : "checkbox",
    "column" : 2,
    "tab" : 3,
    "label" : "structurelewaarde",
    "mandatory" : false,
    "linkedList" : 0,
    "valueTrue" : "Y",
    "valueFalse": "N",
  }

### DIFF
--- a/tailormap-components/projects/core/src/lib/feature-form/form-field/formfield.component.html
+++ b/tailormap-components/projects/core/src/lib/feature-form/form-field/formfield.component.html
@@ -3,6 +3,7 @@
   [value]="humanReadableValue$ | async"
   [label]="attribute.label"
   [fieldType]="attribute.type"
+  [valueTrue]="attribute.value === attribute.valueTrue"
 ></tailormap-label-field>
 
 <tailormap-input-field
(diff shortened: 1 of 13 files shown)


tailormap-components/projects/core/src/lib/feature-form/form-field/formfield.component.html
```diff
@@ -29,3 +30,17 @@
...
+</div>
+
+<div [formGroup]="groep" class="form-field" *ngIf="editing && isDateAttribute(attribute)">
+  <div class="label">{{attribute.label}}</div>
+  <mat-form-field>
+    <input matInput [matDatepicker]="picker" [value]="getDate()" (dateChange)="changeDate($event)" [placeholder]="getDate()">
+    <mat-datepicker-toggle matSuffix [for]="picker"></mat-datepicker-toggle>
+    <mat-datepicker [startAt]="getDate()" #picker></mat-datepicker>
+  </mat-form-field>
+</div>
```
mtoonen: Might be better to move to separate components in user-interface/generic-components. That way we can style form components generically according to the concept7 style

mtoonen (changes requested):
Placing the input types in their own component would simplify the other components. 
Can please write a few lines how this would affect the form configuration json? Two snippets (one for checkboxes, one for the datepicker) with explanations for the new fields should suffice. 
Other than that it's good!

mtoonen (approved):
For now, have to merge, but I think it would be good to have the checkbox also in a separate component

PR #2777, opened 2021-07-14
Added header tailormap-username so the called service knows who sends it

mtoonen:


matthijsln (commented):
I assume this header is meant so endpoints (such as feature-api) can check layer/featuretype authorizations.

Currently only authenticated users are allowed access to proxied backends so the header never has a 'null' value. What about configuring endpoints in the database and allowing access for everyone, all authenticated users, users in group(s)?

mtoonen:
> I assume this header is meant so endpoints (such as feature-api) can check layer/featuretype authorizations.
> 
> Currently only authenticated users are allowed access to proxied backends so the header never has a 'null' value. What about configuring endpoints in the database and allowing access for everyone, all authenticated users, users in group(s)?

This is used in a different API, which stories the user making the request. That API has no principal, so no access to the username.

matthijsln:
> > I assume this header is meant so endpoints (such as feature-api) can check layer/featuretype authorizations.
> > Currently only authenticated users are allowed access to proxied backends so the header never has a 'null' value. What about configuring endpoints in the database and allowing access for everyone, all authenticated users, users in group(s)?
> 
> This is used in a different API, which stories the user making the request. That API has no principal, so no access to the username.

Hence the use of _such as_ :) In the future backend for other purposes may need more fine-grained access control, but that and database config can be added later. As mentioned in #2778, this header is usually called `X-Remote-User` or `X-Forwarded-User` for proxies specifically.

matthijsln:
> > > I assume this header is meant so endpoints (such as feature-api) can check layer/featuretype authorizations.
> > > Currently only authenticated users are allowed access to proxied backends so the header never has a 'null' value. What about configuring endpoints in the database and allowing access for everyone, all authenticated users, users in group(s)?
> > 
> > 
> > This is used in a different API, which stories the user making the request. That API has no principal, so no access to the username.
> 
> Hence the use of _such as_ :) In the future backend for other purposes may need more fine-grained access control, but that and database config can be added later. As mentioned in #2778, this header is usually called `X-Remote-User` or `X-Forwarded-User` for proxies specifically.

mtoonen:
superseded by #2778

PR #2778, opened 2021-07-14
Feat/tm 242 multiple endpoints support

mtoonen:


matthijsln (changes requested):
Get configured endpoints from init parameter or database.

viewer/src/main/java/nl/tailormap/viewer/stripes/ProxyRESTActionBean.java
```diff
@@ -60,6 +66,22 @@ public void checkAuthorization() {
         }
     }
 
+    @Before(stages = LifecycleStage.EventHandling)
+    public void initEndpoints() {
+        endpoints = new HashMap<>();
+        String config = context.getServletContext().getInitParameter("tailormap.restproxy.endpoints");
```
matthijsln, line 72: Why use `0=/path1,1=/path2` and an additional endpoint parameter instead of just using a REST path like `@UrlBinding("/action/restproxy/{endpoint}/{path}")` and not using the `url` parameter at all (and just copying over the original parameters)? I.e. just like an Apache `ProxyPass /action/restproxy/path1/ http://localhost/path1/` would work.

Or if you want to keep the `endpoint` and `url` parameters just use `endpoint=/path1` And just an init parameter without the equals signs, just checking if the `endpoint` parameter matches a configured value?

viewer/src/main/java/nl/tailormap/viewer/stripes/ProxyRESTActionBean.java
```diff
@@ -81,6 +103,7 @@ public void stream(HttpServletResponse response) throws Exception {
         URL theUrl = constructURL();
         HttpClientConfigured client = getHttpClient(theUrl, em);
         HttpUriRequest req = getHttpRequest(theUrl);
+        req.setHeader("tailormap-user", request.getUserPrincipal().getName());
```
matthijsln: This header is usually called `X-Remote-User`

viewer/src/main/java/nl/tailormap/viewer/stripes/ProxyRESTActionBean.java
```diff
@@ -121,7 +145,7 @@ private URL constructURL() throws MalformedURLException {
         URL requestUrl = new URL(context.getRequest().getRequestURL().toString());
         String host = context.getServletContext().getInitParameter("flamingo.restproxy.host");
         String port = context.getServletContext().getInitParameter("flamingo.restproxy.port");
-        String constructedURL = "http://" + (host != null ? host : "localhost") + ":" + (port != null ? port : "8084") + "/feature-api" + url + parentId;
+        String constructedURL = "http://" + (host != null ? host : "localhost") + ":" + (port != null ? port : "8084") + "/" + endpoints.get(endpoint) + url + parentId;
```
matthijsln, line 157: If the `endpoint` parameter is not in the map this will proxy to `http://localhost:8084/null`? Returning 404 seems better

viewer/src/main/webapp/META-INF/context.xml
```diff
@@ -4,6 +4,7 @@
     <Parameter name="tailormap.data.dir" override="false" value="/opt/tailormap_data"/>
     <Parameter name="tailormap.solr.url" override="false" value="https://tailormap.nl/solr/"/>
     <Parameter name="flamingo.restproxy.port" override="false" value="8084"/>
+    <Parameter name="tailormap.restproxy.endpoints" override="false" value="0=feature-api;1=gbi"/>
```
matthijsln: `gbi` is not standard tailormap and should be in a WAR overlay

matthijsln (changes requested):
Also `EntityManager em = Stripersist.getEntityManager();` is not necessary

viewer/src/main/java/nl/tailormap/viewer/stripes/ProxyRESTActionBean.java
mtoonen: More of a preference to use the same concepts (other proxyactionbean), not changing more than needed and the knowledge that this actionbean will be deleted when spring security will be used, so investing time in refactoring this seems like a waste of time

matthijsln (approved):

PR #2822, opened 2021-07-30
[WIP][TM-256] Spring Boot viewer-ng module

mprins:
- [x] setup maven module
- [ ] port `retrieveAppConfigJSON`
- [ ] ...

close TM-256

### DIFF
--- a/viewer-ng/pom.xml
+++ b/viewer-ng/pom.xml
@@ -6,7 +6,7 @@
         <!-- would have org.springframework.bootspring-boot-starter-parent as parent, but we need TM as a parent -->
         <groupId>org.tailormap</groupId>
         <artifactId>tailormap</artifactId>
-        <version>5.9.9-SNAPSHOT</version>
+        <version>5.9.12-SNAPSHOT</version>
     </parent>
     <groupId>org.tailormap</groupId>
     <artifactId>viewer-ng</artifactId>
--- a/viewer-ng/pom.xml
+++ b/viewer-ng/pom.xml
@@ -3,25 +3,19 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
-        <!-- would have org.springframework.bootspring-boot-starter-parent</artifactId> as parent, but we need TM as a parent -->
+        <!-- would have org.springframework.bootspring-boot-starter-parent as parent, but we need TM as a parent -->
         <groupId>org.tailormap</groupId>
         <artifactId>tailormap</artifactId>
         <version>5.9.9-SNAPSHOT</version>
     </parent>
     <groupId>org.tailormap</groupId>
     <artifactId>viewer-ng</artifactId>
-
     <packaging>war</packaging>
     <name>viewer-ng</name>
     <description>Springboot based viewer module</description>
     <properties>
 
     </properties>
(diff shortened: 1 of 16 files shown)


viewer-ng/pom.xml
```diff
@@ -0,0 +1,115 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <!-- would have org.springframework.bootspring-boot-starter-parent</artifactId> as parent, but we need TM as a parent -->
```
mtoonen: Possibly dumb question: using the springboot starter as a parent in the tailormap parent module? Maybe for now not possible (everything would get the springboot deps)

viewer-ng/src/test/java/nl/tailormap/viewer_ng/HSQLDBTestProfileJPAConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+        LocalContainerEntityManagerFactoryBean em = new LocalContainerEntityManagerFactoryBean();
+        em.setPersistenceUnitName("viewer-config-hsqldb");
+        em.setDataSource(dataSource());
+
+        return em;
+    }
+
+    @Bean
+    @Profile("test")
+    public DataSource dataSource() {
```
mtoonen, line 47: This is not necessary when having a separate test propertiesfile (which you have) and you add a annotation to the test-class: 
@TestPropertySource(
        locations = "classpath:application-integrationtest.properties").

When using the default propertynames + hsqldb dependency in you pom file, springboot will automatically create the datasource.

viewer-ng/src/main/webapp/META-INF/context.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Context antiResourceLocking="false" path="/viewer">
```
mtoonen: Nice, I couldn't get this to work...

mtoonen (approved):
Some small changes, but overall looks promising

viewer-ng/pom.xml
mprins: springboot starte doesn't really do anything apart from 1 plugin; we're using the upstream parent as a BOM in our parent now, so we get all the right versions, but can still override with a latest version eg. Hibernate

viewer-ng/src/test/java/nl/tailormap/viewer_ng/HSQLDBTestProfileJPAConfiguration.java
mprins: yeah need to figure that out, needed to upgrade to JUnit 5 first (#2827 )

mprins: it turns out I need to override some things that are specified in the persistence.xml; I have something working now that (for each test):

1. creates a database with a unique name
2. loads the test data

so integration/unit tests can run multithreaded

mprins:
moved to https://github.com/B3Partners/tailormap-api

PR #2837, opened 2021-08-10
Add form editor to viewer admin

mtoonen:


viewer-admin/src/main/java/nl/tailormap/viewer/admin/stripes/FormActionBean.java
```diff
@@ -133,7 +133,7 @@ public void makeLists(){
     @DefaultHandler
     @DontValidate
     public Resolution view() {
-        return new ForwardResolution(JSP);
+        return new ForwardResolution(JSP).addParameter("debug",context.getRequest().getParameter("debug"));
```
geertplaisier, line 142: Is this intentional or dev code?

mtoonen: intentional dev code :)
It doesn't do anything if the debug parameter isn't set, but if it is, the parameter is being persisted (and the debug version of the editor is being used instead of the prod version)

geertplaisier (approved):

mtoonen:
Saving and removing works. Maybe some code cleanup should be done

PR #2886, opened 2021-09-03
Bugfixes

geertplaisier:
TM-232: Fixes some edge cases where layernames are used for featuretype
TM-204: Don't load all features on point if there are no selected layers
TM-264: Fix drawing for layers with an alias
TM-265: Show styling for configured geometry type
TM-266: Don't show passport/edit form for features without Formconfiguration
 TM-268: Allow to edit features from attribute list again


mprins (approved):
LGTM... 😨

PR #2897, opened 2021-09-20
Save string value instead of ID when saving field with linked list

geertplaisier:


tailormap-components/projects/core/src/lib/user-interface/generic-components/input-field/input-field.component.ts
```diff
@@ -18,7 +18,13 @@ export class InputFieldComponent extends BaseFieldComponent implements OnInit {
   }
 
   public getInputType() {
-    return this.fieldType === FormFieldType.HYPERLINK ? 'url' : this.fieldType === FormFieldType.CHECKBOX ? 'checkbox' : 'text';
+    if (this.fieldType === FormFieldType.HIDDEN) {
```
mvdstruijk, line 21: Why is the option checkbox deleted?

mvdstruijk (approved):

PR #2939, opened 2021-10-19
fix LGTM viewer-admin build by skipping the frontend-maven-plugin execution

mprins:
LGTM only checks Java code when running Maven anyways


mprins:
LGTM base commit build failure is expected (as that is what we try to fix here)

PR #2950, opened 2021-10-28
GBIM-220: Allow to download related features with filter on layer

geertplaisier:


mprins (approved):
LGTM, afgezien van de nu overtollige `HttpParams` import in tailormap-components/projects/core/src/lib/shared/export-service/export.service.ts

PR #2970, opened 2021-11-12
GBIM-230: Fix copying nested related features

geertplaisier:
Based on #2961 and #2962 so first review and merge those PR's

Fixes GBIM-230


mprins:
@geertplaisier Er is een merge conflict in deze PR

mprins (approved):

PR #3068, opened 2022-01-27
[HTM-124] make it possible to store the service capabilities document

mprins:
use lazy fetching for the capabilities document, see also https://docs.jboss.org/hibernate/orm/5.6/userguide/html_single/Hibernate_User_Guide.html#BytecodeEnhancement

PR #3072, opened 2022-01-28
[HTM-122] Store WMTS capabilities document when loading or updating service

mprins:
- [x] Store WMTS capabilities document

resolves HTM-122

depends on #3068


mprins:
❗  noting that tiled service in general cannot be updated in tailormap

PR #3205, opened 2022-04-15
[Backport v5.8.x] Replace log4j 1.x with reload4j and remove ancient `eu.medsea.mimeutil:mime-util` dependency

mprins:
backports #3204


mprins:
note that some build faiures are explaned with this PR: https://github.com/actions/toolkit/pull/1015 and this issue: https://github.com/actions/runner/issues/449

there is an `Error: EACCES: permission denied, scandir '/home/runner/work/tailormap/tailormap/viewer-commons/docker/geoserver/geoserver-data/gwc/tmp'` while recomputing the hash

PR #3509, opened 2023-07-28
Use new PDOK api for location search

mvdstruijk:


viewer/src/main/java/nl/b3p/viewer/search/PDOKSearchClient.java
```diff
@@ -43,10 +43,10 @@
  */
 public class PDOKSearchClient extends SearchClient {
 
-    private static final Log log = LogFactory.getLog(SolrSearchClient.class);  
-    private SolrServer server;
+    private static final Log log = LogFactory.getLog(SolrSearchClient.class);
```
mprins, line 46: this results in logging with the wrong handle, `PDOKSearchClient.class` should be used

mprins: see https://github.com/B3Partners/tailormap/pull/3511

